PR #275, opened 2023-12-01
RetroPGF Calculator Checks

MSilb7:
PRing my checks on the RPGF notebook. Feel free to just copy stuff over you want to your branch, then close @chuxinh.

Notes:
- `dummy_data_generator`
  - Modified the dummy votes file generator such that `projects_in_ballot` counts `votes` rather than being a random number

- `rpgf3_allocator`
  - Added some additional logging things and print outs.
    - I like some of the logging adds (i.e. check total amount of OP)
    - The print out / displays are more just to track a specific voter or project through out, to make sure it's correct.
      - ^ We will not publish these when we run actual votes to protect voters.

- `utils`
  - Added an option for a new vote scaling function. The old version ~may overcut projects that fall below the minimum OP threshold (i.e. what if one has 1499 and one has 100 - cutting the 100 likely makes the 1499 one eligible).
    - This new version just peels one-by-one from the bottom and re-scales each time. We likely should increase the max iterations & check on it when we run the script on actual votes.


chuxinh:
Thanks for this Michael, making some more changes on your branch and will merge it:
`dummy_data_generator`:
- Adding a random seed to make sure it's always outputting the same set of addresses and votes 
- Creating logic that if it's `has_published` then `has_voted` is always true, else it's random choice between true or false
- make sure `published_at` time range is between (updated_at, end_date)
- updated `projects_in_ballot`, somehow `count(votes.len())` is not working for me

`utils`:
- to cut amount only those less than (< 1500) 
- update try excep to use if else logic 

`test`:
- create a new test for the new oneby function to see if it works as expected
- we still don't get to see how it works for the iteration tho since it's not what the function itself is doing

PR #282, opened 2023-12-11
Delete Initial File

Billy19191:
Unused python script


chuxinh:
Hi @Billy19191 thanks for making a PR to our repo. 

The `__init__.py` file is to initialise a Python package. In this case as it also needs to import code from other directories within the project, having an `__init__.py` file can help structure your directories as packages and simplify imports.

chuxinh:
Hi @Billy19191 thanks for making a PR to our repo.

The `__init__.py` file is to initialise a Python package. In this case as it also needs to import code from other directories within the project, having an `__init__.py` file can help structure your directories as packages and simplify imports.

PR #351, opened 2024-02-07
Added growthepie origin_key column

mseidlx:
Added column 'growthepie_origin_key' which is the chain identifier that we use in our backend as well as our main API endpoints (api.growthepie.xyz/v1/fundamentals.json and api.growthepie.xyz/v1/contracts.json)

Had some trouble with the CSV (classic date locale issues....) maybe a different file type (i.e. JSON?) is better suited for collaboration


MSilb7:
I realized we were never pulling the 'origin key' out of our GTP API pulls. Updated the helper function to pull this + replaced the alias column with the origin key column (will update downstream queries). Thanks!

PR #668, opened 2024-08-13
Add lisk and swan

chuxinh:
- add list and swan mainnets 
- update the template 

Please have a look @MSilb7 and the pipeline is already running


chuxinh:
also what's the best way to add them into your superchain aggregate logic? @MSilb7

MSilb7:
great, yeah i see the pipelines are running in goldksy now too. to add them to superchain logic for transactional data, we need to update the [chain metadata file](https://github.com/ethereum-optimism/op-analytics/blob/main/op_chains_tracking/inputs/chain_metadata_raw.csv) to:

- add a row if doesn't already exist (i think both do)
- tag as a standard or non-standard config chain in `chain_type`
- add oplabsdb_schema & other mappings (growthepie, defillama, etc) if they exist yet. If not, we just need to stay on top of it / figure out some other process.

MSilb7:
We also need to know Swan's L1 contracts, is that documented anywhere?

MSilb7:
> We also need to know Swan's L1 contracts, DA layer, state layer. We have no info, is that documented anywhere? @chuxinh

Found them on L2Beat. We should make the input form more public / link this somewhere that we show which chains propogate through to the Superchain health dashboard & how they can update themselves.

MSilb7 (approved):
Left comments above, should be good now. We still have a discrepancy on Swan's gas token, but we'll backfill whatever is correct later.

PR #810, opened 2024-09-17
Remove converted .py file and update ipynb so results are not written to txt (already going to BQ)

lithium323:
**Description**

Remove converted .py file and update ipynb so results are not written to txt (already going to BQ). I believe the txt files are no longer being used anywhere else in the repo.


MSilb7:
Hi, the use case for the .txt files is so we can read the latest gas prices in to google sheets for public chain parameter calculators.
- [Google Sheet](https://docs.google.com/spreadsheets/d/1V3CWpeUzXv5Iopw8lBSS8tWoSzyR4PDDwV9cu2kKOrs/edit?gid=186414307#gid=186414307) | [Docs Reference](https://docs.optimism.io/builders/chain-operators/management/blobs#determine-scalar-values-for-using-blobs)
- Formula to call this txt file is `=IMPORTDATA("https://static.optimism.io/op-analytics/reference_data/market_data/outputs/suggest_base_fee.txt")`

Open to other ideas, but we followed this pattern, since FND also does it to create a circulating OP token supply endpoint (but that only updates monthly).

But to reduce commits, maybe we can either:
1. Make this update once per day or so, rather than every hour.
or 2. Push things like this in to a separate repo, so that it doesn't clog things.

lithium323:
> Hi, the use case for the .txt files is so we can read the latest gas prices in to google sheets for public chain parameter calculators.
> 
> * [Google Sheet](https://docs.google.com/spreadsheets/d/1V3CWpeUzXv5Iopw8lBSS8tWoSzyR4PDDwV9cu2kKOrs/edit?gid=186414307#gid=186414307) | [Docs Reference](https://docs.optimism.io/builders/chain-operators/management/blobs#determine-scalar-values-for-using-blobs)
> * Formula to call this txt file is `=IMPORTDATA("https://static.optimism.io/op-analytics/reference_data/market_data/outputs/suggest_base_fee.txt")`
> 
> Open to other ideas, but we followed this pattern, since FND also does it to create a circulating OP token supply endpoint (but that only updates monthly).
> 
> But to reduce commits, maybe we can either: 1 Make this update once per day or so, rather than every hour. or 2 Push things like this in to a separate repo, so that it doesn't clog things.

ah that's good to know! To make the dependency more explicit something we could do is have github action update a specific cell in the destination spreadsheet.  If that sounds good I can look into it.

PR #844, opened 2024-10-01
Added RPGF5 folder and testing scripts

baokiddy:
**Description**

Creating a folder to hold RPGF testing scripts and inputs


lithium323 (approved):
on the json file name I would recommend removing the spaces from the file name. It has ` (2)` on the name.

PR #919, opened 2024-10-23
Fix bug in l2beat data pull: last slug was being assigned to all projects

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**


MSilb7:
confirmed the issue, looks good 
image

MSilb7 (approved):

PR #896, opened 2024-10-16
Add `daily_address_summary` concept model

chuxinh:
Creating some script and logic as per [this sheet](https://docs.google.com/spreadsheets/d/1Ih5GCoN2yLvIYQucJuHVB_f9rFBeqKM7M6rmX3Q7ugM/edit?gid=0#gid=0) during our offsite earlier this month.

The table is grouped by the following:
- adress
- chain_id
- chain
- dt

To dos:
- [ ] Test run on greater time / block range
- [x] Write tests

Any other features that can be added but missing? Anything redundant?

PR #920, opened 2024-10-23
Daily Address Summary model using DuckDB

lithium323:
Rework of #896 in DuckDB


dioptx:
- @chuxinh  Some columns have a denomination suffix ( for example _gwei) while on others it is implied. Since we probably need to denominate certain things on _gwei and others on _eth, could we use suffixes on all of them? An alternative would be to settle on one specific denomination for a given table at least.
- @lithium323 Will materialized queries be accessible somewhere (docs/code) or this is TBD?

lithium323:
> * @lithium323 Will materialized queries be accessible somewhere (docs/code) or this is TBD?

we should have them accessible somewhere, but we haven't defined where or exactly how just yet. Open to suggestions.

chucktraverse:
@lithium323 apologies for being late to the party, but I'm curious about why we are switching to the Expression() pattern for defining the columns and column logic? I understand what it's doing, but are we creating more complexity and lines of code when traditional SQL logic written directly into the query string would suffice?

I can see this getting difficult to review and troubleshoot more complicated queries in the future if so much is abstracted and pulled apart

lithium323:
> @lithium323 apologies for being late to the party, but I'm curious about why we are switching to the Expression() pattern for defining the columns and column logic? I understand what it's doing, but are we creating more complexity and lines of code when traditional SQL logic written directly into the query string would suffice?
> 
> I can see this getting difficult to review and troubleshoot more complicated queries in the future if so much is abstracted and pulled apart

Hi Chuck, thanks for taking a look. Yeah that's a really good point and the kind of feedback we were looking for. You are right, expression is not doing much, only separating the alias from each aggregate expression.  

As I was dissecting the query to understanding it better myself I used this as a way to focus on each individual thing that the query was doing. So it helped me then, but I agree with you in that there's nothing like looking at the final SQL regardless of how you are producing it. 

We can take a stab at producing another copy of this PR with straight up SQL, maybe only some string interpolation. And see how that looks.

chucktraverse:
There's probably a nice middle ground we can find. I like your unit conversion macro and if we can work that in then great!

chuxinh:
I personally don't dislike the way of using Expression here despite it does not feel intuitive to me at the beginning. It also makes the final output of SQL query extremely neat as you can probably tell the reuse of many SQL logics back and forth, especially related to calculating the `ESTIMATED_SIZE` of L1 gas used and other weighted average calcuations.

In terms of Dennis' question:
>@chuxinh Some columns have a denomination suffix ( for example _gwei) while on others it is implied. Since we probably need to denominate certain things on _gwei and others on _eth, could we use suffixes on all of them? An alternative would be to settle on one specific denomination for a given table at least.

By default assuming gas fees are in `_eth` and prices are in `_gwei` and the reason I didn't created suffix for fees in `eth` is because of alternative gas tokens that might emerge in the future? So it's sort of implied there, which I think can be improved in documentation in some way.

lithium323:
Hey everyone, I'm going to go ahead and merge the PR and then we can continue working on follow ups using new PRs.  

Since the processing is not being scheduled yet then no harm in merging, and then future PRs can focus on the aspects that are still under open discussion. 

Expect us to iterate more on this!

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -1,27 +1,34 @@
+# -*- coding: utf-8 -*-
 import duckdb
+from op_coreutils.duckdb_inmem import init_client
 
 from op_datasets.etl.intermediate.registry import register_model
 from op_datasets.etl.intermediate.task import IntermediateModelsTask
-from op_datasets.etl.intermediate.udfs import Expression, wei_to_eth, wei_to_gwei, safe_div, to_sql
+from op_datasets.etl.intermediate.udfs import (
+    Expression,
+    wei_to_eth,
```
MSilb7, line 9: it is embarassing how happy these make me.

alternatively / additionally. I wonder if we can just pull things from [web3.py](https://web3py.readthedocs.io/en/stable/)? It has similar conversion functions & I think also has abi decoding for the future

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -85,66 +112,60 @@
...
+        alias="l2_contrib_gas_fees_basefee",
+        sql_expr=wei_to_eth(f"SUM({L2_CONTRIB_BASE})"),
     ),
     Expression(
-        alias="l2_contrib_gas_fees_priority_fees",
+        alias="l2_contrib_gas_fees_priorityfee",
         sql_expr=wei_to_eth(f"SUM({L2_CONTRIB_PRIORITY})"),
     ),
+    Expression(
+        alias="l2_contrib_gas_fees_legacyfee",
```
MSilb7, line 158: ok got it, we handle the legacyfee here, instead of as a definition. open to either strategy

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -171,63 +192,51 @@
         alias="avg_l1_gas_price_gwei",
         sql_expr=wei_to_gwei(
             safe_div(
-                "SUM(receipt_l1_gas_price * receipt_l1_gas_used)",
-                "SUM(receipt_l1_gas_used)",
+                f"SUM({L1_CONTRIB_L1_GAS})",
+                f"SUM(({ESTIMATED_SIZE}) * COALESCE(16*receipt_l1_base_fee_scalar/1000000, receipt_l1_fee_scalar))",
```
MSilb7, line 196: i wonder if this can be replaced by a standard variable function?

Like, this is `L1_Contrib_L1_Gas/receipt_l1_gas_price` - which also makes it easier to communicate that this is a weighted average.

Similar idea for `L1_CONTRIB_BLOB/receipt_l1_blob_base_fee`

MSilb7 (commented):
🦆 Some comments on continued simplification, but overall exciting

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -1,27 +1,34 @@
...
+from op_datasets.etl.intermediate.udfs import (
+    Expression,
+    wei_to_eth,
+    wei_to_gwei,
+    safe_div,
+    to_sql,
+)
 
 
 # Reused expressions
```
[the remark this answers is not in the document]
chuxinh: Yea I think eventually these can be moved somewhere that we can just use in other queries and logic such as the one you want to build for chain level aggregation

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -85,66 +112,60 @@
...
         alias="total_l1_gas_used_success",
-        sql_expr=f"SUM(IF({TX_SUCCESS}, receipt_l1_gas_used, 0))",
-    ),
-    Expression(
-        alias="total_l1_gas_used_fail",
-        sql_expr=f"SUM(IF({TX_FAIL}, receipt_l1_gas_used, 0))",
+        sql_expr=f"SUM(IF({TX_SUCCESS}, COALESCE(receipt_l1_gas_used, ({ESTIMATED_SIZE})), 0))",
     ),
     # Gas Fee Paid
     Expression(
```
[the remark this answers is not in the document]
chuxinh: Dennis also made a comment before about this. I made it generic as fees to maybe customise for gas tokens, but i wonder if later on we need a gas token field as well 🤷

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
chuxinh: Not sure if i fully get this, but i suppose abstracting away the denominator into something would help readability here

PR #931, opened 2024-10-29
Add functionality & tests to data handling utils

dioptx:
**Description**

- Updated `write.py` in `op_coreutils/bigquery` with:
  - New functions: `write_df_to_bq`, `upsert_partition`, refactored `overwrite_partitions`.
  - Enhanced docstrings and exception handling.
  - `upsert_partition` for unique-key-based data upserts.

- Added test suite `test_write.py` covering:
  - Initialization, overwrite, upsert, and partition handling.
  - Edge cases for empty DataFrames and unique keys in upserts.

**Tests**

- Created tests in `test_write.py` for BigQuery write operations:
  - Valid and invalid date partitioning.
  - Upsert with various unique keys.
  - Error handling for empty DataFrames and invalid configurations.

### DIFF
--- a/packages/op-coreutils/src/op_coreutils/bigquery/write.py
+++ b/packages/op-coreutils/src/op_coreutils/bigquery/write.py
@@ -30,7 +30,7 @@ def init_client():
             _CLIENT = MagicMock()
 
     # In non-PROD environments outside of tests _CLIENT remains uninitialized
-    # so it raises an exception if we try to use it. 
+    # so it raises an exception if we try to use it.
     if _CLIENT is None:
         raise RuntimeError("BigQuery client was not properly initialized.")
 
@@ -60,14 +60,14 @@ def ensure_valid_dt(df: pl.DataFrame, dt=None):
     if dt is None:
         dt_values = df["dt"].unique().to_list()
         if not dt_values:
-            raise OPLabsBigQueryError("No valid 'dt' values found in DataFrame.")
+            log.info("No valid 'dt' values found in DataFrame.")
+            return
         dt = str(dt_values[0])
 
     if isinstance(dt, datetime) or hasattr(dt, "strftime"):
         dt = dt.strftime("%Y-%m-%d")
     elif isinstance(dt, str):
         dt = dt.split(" ")[0]
-
     try:
         datetime.strptime(dt, "%Y-%m-%d")
     except Exception as ex:
--- a/packages/op-coreutils/src/op_coreutils/bigquery/write.py
+++ b/packages/op-coreutils/src/op_coreutils/bigquery/write.py
@@ -80,7 +80,7 @@ def write_df_to_bq(
     destination: str,
     client: bigquery.Client,
     operation: str,
-    write_disposition: str = bigquery.WriteDisposition.WRITE_TRUNCATE,
+    write_disposition: str,
     time_partitioning: bigquery.TimePartitioning = None,
 ):
     """Helper function to write a DataFrame to BigQuery.
@@ -90,7 +90,7 @@ def write_df_to_bq(
         destination (str): The BigQuery table destination in 'dataset.table' format.
         client (bigquery.Client): The BigQuery client.
         operation (str): Description of the operation for logging.
-        write_disposition (str, optional): BigQuery write disposition. Defaults to 'WRITE_TRUNCATE'.
+        write_disposition (str, optional): BigQuery write disposition.
         time_partitioning (bigquery.TimePartitioning, optional): Time partitioning configuration.
     """
     with io.BytesIO() as stream:
--- a/packages/op-coreutils/src/op_coreutils/bigquery/write.py
+++ b/packages/op-coreutils/src/op_coreutils/bigquery/write.py
@@ -1,111 +1,169 @@
 # -*- coding: utf-8 -*-
 import io
-
-import polars as pl
-
+import sys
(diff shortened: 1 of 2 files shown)


packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -1,150 +1,235 @@
...
+        operation="OVERWRITE PARTITION",
+        time_partitioning=time_partitioning,
+    )
 
-        log.info(
-            f"{operation}OVERWRITE PARTITION: Wrote {human_rows(len(df))} {human_size(filesize)} to BQ {destination}"
-        )
+
+def upsert_partition(
+    df: pl.DataFrame, dt: str, dataset: str, table_name: str, unique_keys=["dt"]
```
MSilb7, line 194: so the default case is equivalent to the dt partition, but we can specify different partitions if/when needed?

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -1,150 +1,235 @@
...
+    """
+    client = init_client()
+    ensure_valid_dt(df, dt)
+    df = df.with_columns(dt=pl.lit(dt).str.strptime(pl.Datetime, "%Y-%m-%d"))
+
+    destination = f"{dataset}.{table_name}"
+    staging_table_name = f"{table_name}_staging"
+    staging_destination = f"{dataset}.{staging_table_name}"
+
+    write_df_to_bq(df, staging_destination, client, operation="WRITE STAGING TABLE")
```
MSilb7, line 213: One note - I got errors in the past when writing to an existing `_staging` table (the table existed if a prior run errored out in the middle).

So I wonder if this needs a step to first delete the staging table if it exists?

MSilb7 (approved):
Upsert seems fine, but I wonder if we're comfortable with how many changes were made to the existing scripts? I'll leave that up to @lithium323

tests/op_coreutils/bigquery/test_write.py
```diff
@@ -0,0 +1,260 @@
...
+def reset_client():
+    """Utility function to reset the global _CLIENT variable."""
+    import op_coreutils.bigquery.write
+
+    op_coreutils.bigquery.write._CLIENT = None
+
+
+@patch("op_coreutils.bigquery.write.get_credentials")
+@patch("op_coreutils.bigquery.write.bigquery.Client")
+def test_init_client(mock_bigquery_client, mock_get_credentials):
```
lithium323, line 49: I'm not sure testing the initialization method is warranted. It seems that for the most part, the test is checking that the mocks where set up correctly. Maybe consider removing this test?

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -1,150 +1,235 @@
...
 _CLIENT: bigquery.Client | MagicMock | None = None
 
 
 def init_client():
-    """Idempotent env-aware client initialization.
-
-    - Guarantess only one global instance exists.
-    - Uses a mock when not running in PROD.
-    """
+    """BigQuery client with environment and testing awareness."""
```
lithium323, line 20: nice, like the conciseness!

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -1,150 +1,235 @@
...
+
+    Args:
+        df (pl.DataFrame): The DataFrame to validate.
+        dt (str, optional): The date string to validate. If None, uses the first unique 'dt' value from the DataFrame.
+
+    Raises:
+        ValueError: If the DataFrame is empty or missing 'dt' column.
+        OPLabsBigQueryError: If 'dt' is not a valid date in 'YYYY-MM-DD' format.
+    """
+    if df.is_empty() or "dt" not in df.columns:
```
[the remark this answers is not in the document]
lithium323: Maybe we should get on the same page as to what the purpose of `ensure_valid_dt`.  My initial though was that it would be used when manually passing a string to be used as the value for `dt`, which would be done on cases where the dataframe did not have a `dt` column. 

But maybe my original premise is wrong altogether. Instead of passing `dt` as a string when not present in the dataframe, if we were to pass it as a `date` object then `ensure_valid_dt` wouldn't be needed at all. 

Curious what you think.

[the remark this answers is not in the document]
dioptx: In my experience `date` objects perform better since they can hold timezone metadata, and that's where a lot of `datetime` related bugs occur.

Now on `ensure_valid_dt`:
If we want to standardize passing `date` objects on an upstream layer (which I think is best to do it somewhere centralized and not on the gcs interfaces) then it is not needed. 

I will remove the enforcement (`if df.is_empty() or "dt" not in df.columns:`) to start with.

packages/op-coreutils/src/op_coreutils/bigquery/write.py
dioptx: Correct, you can specify two keys for example

packages/op-coreutils/src/op_coreutils/bigquery/write.py
dioptx: Great point! To ensure the function handles this, I will add step to delete the staging table if it exists already.

PR #937, opened 2024-10-30
Create helper class for intermediate model test cases. Use it for the `daily_address_summary` model

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
+++ b/packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
@@ -1,9 +1,8 @@
 # -*- coding: utf-8 -*-
 import duckdb
-from op_coreutils.duckdb_inmem import init_client
 
 from op_datasets.etl.intermediate.registry import register_model
-from op_datasets.etl.intermediate.task import IntermediateModelsTask
+from op_datasets.etl.intermediate.types import NamedRelations
 from op_datasets.etl.intermediate.udfs import (
     Expression,
     wei_to_eth,
@@ -27,7 +26,9 @@ L2_CONTRIB_BASE = "(base_fee_per_gas * receipt_gas_used)"
 
 ESTIMATED_SIZE = "receipt_l1_fee /(16*COALESCE(receipt_l1_fee_scalar,receipt_l1_base_fee_scalar)*receipt_l1_gas_price/1000000 + COALESCE( receipt_l1_blob_base_fee_scalar*receipt_l1_blob_base_fee/1000000 , 0))"
 
-L1_CONTRIB_BLOB = f"({ESTIMATED_SIZE}) * receipt_l1_blob_base_fee_scalar/1000000 * receipt_l1_blob_base_fee"
+L1_CONTRIB_BLOB = (
+    f"({ESTIMATED_SIZE}) * receipt_l1_blob_base_fee_scalar/1000000 * receipt_l1_blob_base_fee"
+)
 L1_CONTRIB_L1_GAS = f"({ESTIMATED_SIZE}) * COALESCE(16*receipt_l1_base_fee_scalar/1000000, receipt_l1_fee_scalar) * receipt_l1_gas_price"
 
 
@@ -39,7 +40,7 @@ AGGREGATION_EXPRS = [
     ),
     Expression(
         alias="total_txs_success",
-        sql_expr=f"COUNT_IF({TX_SUCCESS})",
+        sql_expr=f"COUNT(IF({TX_SUCCESS}, 1, NULL))",
     ),
     # Blocks
     Expression(
@@ -211,19 +212,18 @@ AGGREGATION_EXPRS = [
 
 @register_model
 def daily_address_summary(
-    task: IntermediateModelsTask,
-) -> dict[str, duckdb.DuckDBPyRelation]:
-    txs: duckdb.DuckDBPyRelation = task.input_duckdb_relations["transactions"]
-    blks: duckdb.DuckDBPyRelation = task.input_duckdb_relations["blocks"]
-
-    client = init_client()
+    duckdb_client: duckdb.DuckDBPyConnection,
+    input_tables: NamedRelations,
+) -> NamedRelations:
+    input_transactions: duckdb.DuckDBPyRelation = input_tables["transactions"]
+    input_blocks: duckdb.DuckDBPyRelation = input_tables["blocks"]
 
     query = f"""
     WITH blocks AS (
         SELECT
             number,
             base_fee_per_gas
-        FROM blks
+        FROM input_blocks
     )
     SELECT
         dt,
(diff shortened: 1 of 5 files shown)


packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -39,7 +40,7 @@
     ),
     Expression(
         alias="total_txs_success",
-        sql_expr=f"COUNT_IF({TX_SUCCESS})",
+        sql_expr=f"COUNT(IF({TX_SUCCESS}, 1, NULL))",
```
lithium323, line 43: This changed because using COUNT_IF was returning a HUGEINT type, whereas COUNT returns a BIGINT. We want the output schema to be BIGINT.

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
```diff
@@ -211,27 +212,26 @@
 
 @register_model
 def daily_address_summary(
-    task: IntermediateModelsTask,
-) -> dict[str, duckdb.DuckDBPyRelation]:
-    txs: duckdb.DuckDBPyRelation = task.input_duckdb_relations["transactions"]
-    blks: duckdb.DuckDBPyRelation = task.input_duckdb_relations["blocks"]
-
-    client = init_client()
+    duckdb_client: duckdb.DuckDBPyConnection,
```
lithium323, line 215: slight change, but model functions now require that the duckdb_client is passed into them. This makes it easier to use them in tests.

tests/op_datasets/etl/intermediate/models/test_daily_address_summary.py
```diff
@@ -1,108 +1,322 @@
...
+            "min_block_timestamp": "UINTEGER",
+            "max_block_timestamp": "UINTEGER",
+            "time_interval_active": "UINTEGER",
+            "unique_hours_active": "BIGINT",
+            "num_to_addresses": "BIGINT",
+            "num_to_addresses_success": "BIGINT",
+            "num_method_ids": "BIGINT",
+            "total_l2_gas_used": "DECIMAL(38,0)",
+            "total_l2_gas_used_success": "DECIMAL(38,0)",
+            # TODO: Should l1_gas_used values be DECIMAL(38,0) instead of DOUBLE ???
```
lithium323, line 86: @chuxinh  I'll need your input on this question. Using the current model definition the types for `total_l1_gas_used` and `total_l1_gas_used_success` are coming out as `DOUBLE`.  I wonder if we need to update the definition to make sure it comes out as decimal or bigint.

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
chuxinh: Does this structure apply to future intermediate tables?

packages/op-datasets/src/op_datasets/etl/intermediate/testutils.py
```diff
@@ -0,0 +1,170 @@
...
+
+            else:
+                log.info("Fetching test data from GCS.")
+                cls._fetch_test_data()
+                log.info("Fetched test data from GCS.")
+        else:
+            log.info(f"Using local test data from: {db_path}")
+
+        log.info("Executing model...")
+        model = REGISTERED_INTERMEDIATE_MODELS["daily_address_summary"]
```
chuxinh, line 100: should the model to be tested as a variable?

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
lithium323: Yes! for new models the function signature will have those two parameters.  There is also now a requirement to be more clear about what data does a model consume. You don't see it yet in this PR, but in the main branch it now looks like this:
```
@register_model(
    input_datasets=["blocks", "transactions"],
    expected_outputs=["daily_address_summary_v1"],
)
def daily_address_summary(
    duckdb_client: duckdb.DuckDBPyConnection,
    input_tables: NamedRelations,
) -> NamedRelations:
```

So here the `daily_address_summary` function is producing results that will be stored in gcs under a versioned name: `daily_address_summary_v1`.  This will allow us to decouple the name of the function from where we want to  store the results.

packages/op-datasets/src/op_datasets/etl/intermediate/testutils.py
lithium323: yeah, this helps make sure that the model function is correctly registered with that name.  When running the model transformation in production the command line invocation uses that same string. For example:
```
$ uv run python opdata chains intermediate_models op daily_address_summary @20241001:+1 --read_from gcs --write-to gcs
```

packages/op-datasets/src/op_datasets/etl/intermediate/models/daily_address_summary.py
chuxinh: Would like to see some results when it's available in some test run whenever

lithium323: definitely, working on that as we speak, hoping to show you some data in an actual table by tomorrow

tests/op_datasets/etl/intermediate/models/test_daily_address_summary.py
```diff
@@ -1,108 +1,322 @@
...
-    )
-    assert pl.select(df_multi_txs["l1_contrib_gas_fees"]).item() == Decimal(
-        str(0.0000001497346364060)
-    )
+INPUT_TEST_DATA = InputTestData.at(__file__)
+
+
+class TestDailyAddressSummary001(IntermediateModelTestBase):
+    model = "daily_address_summary"
+    inputdata = InputTestData.at(__file__)
```
chuxinh, line 17: should this just be the `INPUT_TEST_DATA`?

tests/op_datasets/etl/intermediate/models/test_daily_address_summary.py
chuxinh: oh maybe because it's calculated that's why. I think we should keep it consistent as `DECIMAL(38, 0)`

lithium323: cool, then maybe there is something we have to adjust in the expression so that it never coerces to a DOUBLE. My guess is that the division operations are causing things to become double.  Instead of dividing one can instead multiply by the equivalent DECIMAL value.

tests/op_datasets/etl/intermediate/models/test_daily_address_summary.py
```diff
@@ -1,108 +1,322 @@
...
+        "{block_number} = 126073799",
+        "{block_number} = 126090026",
+        "{block_number} % 1000 <= 2",
+    ]
+
+    _enable_fetching = True
+
+    def test_system_address_not_included(self):
+        ans = self._duckdb_client.sql(f"""
+        SELECT * FROM daily_address_summary WHERE address = '{SYSTEM_ADDRESS}'
```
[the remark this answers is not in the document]
lithium323: yeah that's right, the results of the model are stored in a temporary table that can be used in queries from the `self._duckdb_client`.  This lets us write the tests as SQL queries also.

PR #944, opened 2024-10-31
Add upsert_partition functionality & testing for write.py functions

dioptx:
**Description**

- Creared an `upsert_partition` function
- Created tests for the new function as well as the existing ones


packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -134,7 +138,12 @@ def overwrite_partition_static(
     )
 
 
-def _write_df_to_bq(df: pl.DataFrame, destination: str, job_config=bigquery.LoadJobConfig):
+def _write_df_to_bq(
+    df: pl.DataFrame,
+    destination: str,
+    job_config=bigquery.LoadJobConfig,
+    operation_prefix: str = "",
```
lithium323, line 145: it looks like the newly added `operation_prefix` arg is not being used. Should it be removed?

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -152,3 +161,55 @@ def _write_df_to_bq(df: pl.DataFrame, destination: str, job_config=bigquery.Load
...
+
+    df = df.with_columns(dt=pl.lit(dt).str.strptime(pl.Datetime, "%Y-%m-%d"))
+
+    destination = f"{dataset}.{table_name}"
+    staging_table_name = f"{table_name}_staging"
+    staging_destination = f"{dataset}.{staging_table_name}"
+
+    _write_df_to_bq(
+        df,
+        staging_destination,
```
lithium323, line 188: when we do upserts, should we have a dedicated dataset where the staging tables go? That way we won't be mixing production tables with temporary tables used in upserts.

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -152,3 +161,55 @@ def _write_df_to_bq(df: pl.DataFrame, destination: str, job_config=bigquery.Load
...
+        job_config=bigquery.LoadJobConfig(
```
lithium323, line 189: For the staging table, we should use the WRITE_EMPTY write disposition. That way we make sure we are not colliding with something else possibly going on at the same time. 

For the same reason it might also be good to generate a small random suffix to use on the staging table name. That way the staging table will be guaranteed to be unique to this upsert call. 

We should also add an expiration time on the staging table, something like 30 minutes, just in case something breaks we know the table will be automatically deleted by BQ later on.

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -152,3 +161,55 @@ def _write_df_to_bq(df: pl.DataFrame, destination: str, job_config=bigquery.Load
         log.info(
             f"{operation_prefix}{job_config.write_disposition}: Wrote {human_rows(len(df))} {human_size(filesize)} to BQ {destination}"
         )
+
+
+def upsert_partition(
+    df: pl.DataFrame, dt: str, dataset: str, table_name: str, unique_keys=["dt"]
+):
+    """Upsert data into a specific partition in a BigQuery table.
```
lithium323, line 169: I think there's something to clarify here. Do we expect the upsert method to be used with a partitioned table always? From the name of the function it sounds like it.  

Also what would be the expectations on the incoming df, should it have only 1 `dt` value or will the function deal with it possibly containing several `dt` values?

tests/op_coreutils/bigquery/test_write.py
```diff
@@ -0,0 +1,99 @@
...
+        "category": np.random.choice(["A", "B", "C", "D"], num_rows),
+        "value": np.random.uniform(1.0, 10.0, size=num_rows).round(2),
+    }
+)
+
+# Sort by 'dt' for consistency
+test_df = test_df.sort("dt")
+
+
+@patch("op_coreutils.bigquery.write.init_client")
```
lithium323, line 34: We should not need to patch the init_client method. The method itself should be "env aware", so if we are running in tests you will get a MagicMock withou having to patch it in each test yourself.

tests/op_coreutils/bigquery/test_write.py
```diff
@@ -0,0 +1,99 @@
...
+
+    mock_load_job = MagicMock()
+    mock_client.load_table_from_file.return_value = mock_load_job
+
+    overwrite_table(test_df, "test_dataset", "test_table_staging")
+    args, kwargs = mock_client.load_table_from_file.call_args
+    assert kwargs["destination"] == "test_dataset.test_table_staging"
+    assert kwargs["job_config"].write_disposition == "WRITE_TRUNCATE"
+
+    # Testing with incorrect table name to raise custom error
```
lithium323, line 47: nice, helpful to test this constraint!

tests/op_coreutils/bigquery/test_write.py
```diff
@@ -0,0 +1,99 @@
...
+    with pytest.raises(
+        OPLabsBigQueryError, match="cannot overwrite data at test_dataset.test_table"
+    ):
+        overwrite_table(test_df, "test_dataset", "test_table")
+
+
+@patch("op_coreutils.bigquery.write.init_client")
```
lithium323, line 54: similar comment as above. Our client should be aware of where its running so that we don't have to patch it on each test. (if it is not already working like that for some reason then we should address that)

For the assertions you can call init_client() to get the MagicMock instance

lithium323 (changes requested):
nice! Left an initial set of questions and comments to be addressed.

packages/op-coreutils/src/op_coreutils/bigquery/write.py
dioptx: > Do we expect the upsert method to be used with a partitioned table always?

Good point, I increased the scope of the function and turned it into `upsert_table`. It should now handle both cases

packages/op-coreutils/src/op_coreutils/bigquery/write.py
dioptx: Good call, added a `"_staging"` suffix paired with a uid to avoid collisions

tests/op_coreutils/bigquery/test_write.py
dioptx: I was pulling the init_client from `packages/op-coreutils/src/op_coreutils/gsheets.py` instead of `packages/op-coreutils/src/op_coreutils/bigquery/write.py`. 
I corrected this throughout the tests and removed the redundant patches.

lithium323:
good morning @dioptx  was just looking at the updated PR and it is missing some of the bug fix updates that I was telling you about yesterday (#947) you'll have to merge main or rebase onto main so this PR has no conflicts.  Sorry for the trouble 🙏

lithium323 (approved):

PR #955, opened 2024-11-01
Add github analytics data pull (uses the upsert functionality created by Dennis) 

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**


lithium323:
> Looks good to me. Just quick confirmation @lithium323, for new repos I just need to add to the list as needed, right?

yep, that's correct!

PR #958, opened 2024-11-02
Add Chain Colors Util

chuxinh:
Let me know if we still want to continue developing in `helper_functions` or elsewhere in the `op-coreutils`

Aim to create some colour mapping based on chain identifier to generate chain colours according to their brand colour specified in the chain metadata


helper_functions/get_chain_colors.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+from op_coreutils.logger import structlog
+import pandas as pd
+
+log = structlog.get_logger()
+
+FILE_PATH = "../op_chains_tracking/outputs/chain_metadata.csv"
+
+
+def get_chain_colors(chain_identifier: str = "chain_name") -> dict[str, str]:
```
lithium323, line 10: at a glance it does look like the chain color functionality belongs in op_datasets, in the  `op_dataset/chains/chain_metadata.py` file.

I would also like to move us over to using Google Sheets as the source of truth for chain metadata (and away from chain_metadata_raw.csv in the repo).  

@MSilb7  we should touch base on this. To clarify what I had in mind I wrote a "Manual-Entry Tabular Source of Truth" section [here](https://docs.google.com/document/d/1n7V7Adg-u50FggeEY4CzT3mHTQ67wp3gn93PdjCFpF0/edit?tab=t.0#heading=h.5r4ya84moo6y). This describes the pattern that I have in mind.

Using the current tooling to support Google Sheets as the source of truth for chain metadata the colors implementation would look like this:
```
def chain_colors(chain_identifier: str = "chain_name"):
    df = load_chain_metadata().select(chain_idenetifier, "hex_color").filter(pl.col("hex_color").is_not_null())

    colors = []
    for chain in df.to_dicts():
        colors.append({
            chain_identifier: chain[chain_identifier],
            hex_color: "#" + chain["hex_color"].strip().upper()
        })
   
    return colors

chuxinh: Thanks Pedro! I've got it moved under `op_dataset/chains/chain_metadata.py` and I want to keep in the original dictionary format since it's more directly useable in plotting such as matplotlib and plotly.

lithium323 (approved):

Missjones2 (commented):
Merge

Issue #887, opened 2024-10-16
Update defillama stablecoins api pull to only write data to bigquery for the most recent N days


lithium323:
To avoid rewriting all history for all stablecoin every day we should start filtering the API results for the most recent N days (e.g. 30 days). 

Also we should allow parameters to run the stablecoins api pull only for a single stablecoin so that we can selectively backfill if a new stablecoin is added by defillama.  For this to work when writing data to bigquery we need to add a utility that can upsert data into a partition in BQ (the repo already has an implementation of this but it is not yet part of op_coreutils).  Upserts are needed so that we can add data from one stablecoin to a partition without affecting data for other stablecoins already present.

The bigquery upserts feature is described in issue #886 .

PR #964, opened 2024-11-04
Updated existing defillama code to address #887 & improve overall functionality

dioptx:
**Description**

Updated `defillama.py` to align with the new `write.py` functionality and resolve #[887](https://github.com/ethereum-optimism/op-analytics/issues/887)

Key changes:

- **Filtered API Results**: Modified data processing to only include the most recent N days, preventing daily rewrites of the entire history.
- **Selective Stablecoin Pull**: Added a `stablecoin_ids` parameter to `pull_stables`, allowing selective data retrieval and backfilling for specific stablecoins.
- **BigQuery Upsert Operations**: Replaced previous write operations with upsert functions (`overwrite_unpartitioned_table`, `overwrite_partition_static`, `upsert_partitioned_table`) to update BigQuery partitions without overwriting existing data.

**Tests**

Updated `test_defillama.py`:

- **Adjusted Mocks**: Updated mocked functions to match the new function names and signatures. Previously the mock data included an entire defillama response .csv which was multiple MBs
- **Verified Function Calls**: Ensured correct parameters are passed to the BigQuery functions, including `partition_dt` and `unique_keys`.
- **Validated Data Processing**: Confirmed that data filtering and processing work as expected with the new code through updated assertions.

### DIFF
--- a/src/op_analytics/cli/subcommands/pulls/defillama.py
+++ b/src/op_analytics/cli/subcommands/pulls/defillama.py
@@ -1,16 +1,17 @@
 # -*- coding: utf-8 -*-
+from datetime import datetime, timedelta, timezone
 from typing import Dict, List, Optional, Tuple
+
 import polars as pl
 from op_coreutils.bigquery.write import (
-    overwrite_unpartitioned_table,
     overwrite_partition_static,
+    overwrite_unpartitioned_table,
     upsert_partitioned_table,
 )
-from datetime import datetime, timedelta, timezone
 from op_coreutils.logger import structlog
-from op_coreutils.request import new_session, get_data
-from op_coreutils.time import now_date
+from op_coreutils.request import get_data, new_session
 from op_coreutils.threads import run_concurrently
+from op_coreutils.time import now_date
 
 log = structlog.get_logger()
 
--- a/src/op_analytics/cli/subcommands/pulls/defillama.py
+++ b/src/op_analytics/cli/subcommands/pulls/defillama.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import time
 from typing import Dict, List, Optional, Tuple
 import polars as pl
 from op_coreutils.bigquery.write import (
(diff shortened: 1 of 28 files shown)


src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,70 @@
...
+    start = time.time()
+    try:
+        response = session.get(url, headers={"Content-Type": "application/json"})
+        response.raise_for_status()
+        resp_json = response.json()
+        log.info(f"Fetched from {url}: {time.time() - start:.2f} seconds")
+        return resp_json
+    except Exception as e:
+        log.error(f"Failed to fetch data from {url}: {e}")
+        return None
```
lithium323, line 47: My recommendation would be not to catch the exception. If there is an error we want to be aware of it.  But am I missing something?

While working on the github analytics api pull I also realized my get_data method was not checking for the http status code of the response. So perhaps we should also check for status = 200 and raise an exception if we don't get 200. 

Since we are mostly doing the same for all http requests we  can move the get_data() function over to `op_coreutils.request` and not redefine the same functionality in `defillama.py` and `l2beat.py` and `github_analytics.py`.  

Let me take an initial stab at the get_data function in the github_anlytics pr (#955 ) and we can then go back and use that function in this one.

lithium323: here is the common version of the data that I think we could use here: https://github.com/ethereum-optimism/op-analytics/blob/main/packages/op-coreutils/src/op_coreutils/request.py#L28

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,70 @@
...
+    Args:
+        data (dict): The breakdown data from the API.
+        days (int): The number of days to filter in the processed data.
 
-            try:
-                row["bridged_to"] = datapoint["bridgedTo"][peg_type]
-            except KeyError:
-                row["bridged_to"] = None
+    Returns:
+        Tuple[pl.DataFrame, Dict[str, Optional[str]]]: A Polars DataFrame with breakdown data and metadata dictionary.
```
lithium323, line 61: Since the function has type hints perhaps it would be ok to leave out types from docstrings for less redundancy.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,70 @@
...
+    """
+    if data is None:
```
lithium323, line 63: My recommendation is to lean towards always failing instead of working around certain kind of failures.

PR #967, opened 2024-11-04
Update pre-commit settings to match ruff autoformat settings

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,5 +11,4 @@ repos:
     hooks:
       - id: check-yaml
       - id: end-of-file-fixer
-      - id: fix-encoding-pragma
       - id: trailing-whitespace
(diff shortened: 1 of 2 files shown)


dioptx (approved):
@lithium323  Related question apart from LGTM:
Since we are swapping black for ruff, should we do an entire run on the repo in a PR just to clean up everything?

lithium323:
> @lithium323 Related question apart from LGTM: Since we are swapping black for ruff, should we do an entire run on the repo in a PR just to clean up everything?

hey, that could be a bit too invasive for things outside of src/ and packages/, so I would hold off on a full run outside of those two directories.  

For those two directories my recommendation would be to wait until we put together the CI build that enforces formatting.  Once that's in place it will tell us exactly what we need to update.  For now not a priority. 

I'm currently doing a big push on the public datasets front, once that's done I'll get to setting up proper CI checks. So probably early next week.

PR #968, opened 2024-11-04
Updated defillama codebase for #887 & improve overall functionality

dioptx:
Description

Updated defillama.py to align with the new write.py functionality and resolve #https://github.com/ethereum-optimism/op-analytics/issues/887

Key changes:

Filtered API Results: Modified data processing to only include the most recent N days, preventing daily rewrites of the entire history.
Selective Stablecoin Pull: Added a stablecoin_ids parameter to pull_stables, allowing selective data retrieval and backfilling for specific stablecoins.
BigQuery Upsert Operations: Replaced previous write operations with upsert functions (overwrite_unpartitioned_table, overwrite_partition_static, upsert_partitioned_table) to update BigQuery partitions without overwriting existing data.
Tests

Updated test_defillama.py:

Adjusted Mocks: Updated mocked functions to match the new function names and signatures. Previously the mock data included an entire defillama response .csv which was multiple MBs
Verified Function Calls: Ensured correct parameters are passed to the BigQuery functions, including partition_dt and unique_keys.
Validated Data Processing: Confirmed that data filtering and processing work as expected with the new code through updated assertions.

### DIFF
--- a/src/op_analytics/cli/subcommands/pulls/defillama.py
+++ b/src/op_analytics/cli/subcommands/pulls/defillama.py
@@ -1,15 +1,17 @@
-import time
+# -*- coding: utf-8 -*-
+from datetime import datetime, timedelta, timezone
+from typing import Dict, List, Optional, Tuple
+
 import polars as pl
 from op_coreutils.bigquery.write import (
-    most_recent_dates,
     overwrite_partition_static,
-    overwrite_partitions_dynamic,
     overwrite_unpartitioned_table,
+    upsert_partitioned_table,
 )
 from op_coreutils.logger import structlog
-from op_coreutils.request import new_session
-from op_coreutils.time import dt_fromepoch, now_date
+from op_coreutils.request import get_data, new_session
 from op_coreutils.threads import run_concurrently
+from op_coreutils.time import now_date
 
 log = structlog.get_logger()
 
@@ -22,50 +24,44 @@ BREAKDOWN_TABLE = "defillama_daily_stablecoins_breakdown"
 METADATA_TABLE = "defillama_stablecoins_metadata"
 
 
-def get_data(session, url):
-    start = time.time()
-    resp = session.request(
-        method="GET",
-        url=url,
-        headers={"Content-Type": "application/json"},
-    ).json()
-    log.info(f"Fetched from {url}: {time.time() - start:.2f} seconds")
-    return resp
+def process_breakdown_stables(
+    data: dict, days: int = 30
+) -> Tuple[pl.DataFrame, Dict[str, Optional[str]]]:
+    """
+    Processes breakdown data for stablecoins, filtering for the most recent N days.
 
+    Args:
+        data: The breakdown data from the API.
+        days: The number of days to filter in the processed data.
 
-def process_breakdown_stables(data):
-    peg_type = data["pegType"]
-    balances = data["chainBalances"]
+    Returns:
+        A tuple containing a Polars DataFrame with breakdown data and a metadata dictionary.
+    """
+    peg_type: str = data.get("pegType", "")
+    balances: Dict[str, dict] = data.get("chainBalances", {})
 
-    rows = []
+    cutoff_date = datetime.now(timezone.utc) - timedelta(days=days)
+    rows: List[Dict[str, Optional[str]]] = []
+
     for chain, balance in balances.items():
-        tokens = balance["tokens"]
+        tokens = balance.get("tokens", [])
+        filtered_tokens = [
+            datapoint
+            for datapoint in tokens
+            if datetime.fromtimestamp(datapoint["date"], tz=timezone.utc) >= cutoff_date
+        ]
 
-        for datapoint in tokens:
-            row = {}
-            row["chain"] = chain
-            row["dt"] = dt_fromepoch(datapoint["date"])
-
-            try:
-                row["circulating"] = datapoint["circulating"][peg_type]
-            except KeyError:
-                row["circulating"] = None
-
-            try:
-                row["bridged_to"] = datapoint["bridgedTo"][peg_type]
-            except KeyError:
-                row["bridged_to"] = None
-
-            try:
-                row["minted"] = datapoint["minted"][peg_type]
-            except KeyError:
-                row["minted"] = None
-
-            try:
-                row["unreleased"] = datapoint["unreleased"][peg_type]
-            except KeyError:
-                row["unreleased"] = None
-
+        for datapoint in filtered_tokens:
+            row: Dict[str, Optional[str]] = {
+                "chain": chain,
+                "dt": datetime.fromtimestamp(
+                    datapoint["date"], tz=timezone.utc
+                ).strftime("%Y-%m-%d"),
+                "circulating": datapoint.get("circulating", {}).get(peg_type),
+                "bridged_to": datapoint.get("bridgedTo", {}).get(peg_type),
+                "minted": datapoint.get("minted", {}).get(peg_type),
+                "unreleased": datapoint.get("unreleased", {}).get(peg_type),
+            }
             rows.append(row)
 
     must_have_metadata_fields = [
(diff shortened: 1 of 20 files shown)


src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,44 @@
...
 
-            try:
-                row["minted"] = datapoint["minted"][peg_type]
-            except KeyError:
-                row["minted"] = None
+    Returns:
+        A tuple containing a Polars DataFrame with breakdown data and a metadata dictionary.
+    """
+    peg_type: str = data.get("pegType", "")
+    balances: Dict[str, dict] = data.get("chainBalances", {})
```
lithium323, line 41: I think we shouldn't have defaults, we want to make sure the data is exactly as we expect it to be otherwise we might end up ingesting some invalid data.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +84,102 @@ def process_breakdown_stables(data):
...
-    for key in metadata_fields:
-        metadata[key] = data.get(key)
-
-    result = pl.DataFrame(rows, infer_schema_length=len(rows)).with_columns(
-        id=pl.lit(metadata["id"]),
-        name=pl.lit(metadata["name"]),
-        symbol=pl.lit(metadata["symbol"]),
+    metadata: Dict[str, Optional[str]] = {
+        key: data.get(key) for key in must_have_metadata_fields + metadata_fields
+    }
```
lithium323, line 89: I think that the list-comprehension is making the code shorter but obscuring the purpose

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,44 @@
...
 
+    for chain, balance in balances.items():
+        tokens = balance.get("tokens", [])
+        filtered_tokens = [
+            datapoint
+            for datapoint in tokens
+            if datetime.fromtimestamp(datapoint["date"], tz=timezone.utc) >= cutoff_date
+        ]
+
+        for datapoint in filtered_tokens:
```
lithium323, line 54:
```suggestion
        from op_coreutils.time import datetime_fromepoch

        for datapoint in tokens:
            if datetime_fromepoch(datapoint["date"]) < cutoff_date:
                continue
```

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,44 @@
...
+            row: Dict[str, Optional[str]] = {
+                "chain": chain,
+                "dt": datetime.fromtimestamp(
+                    datapoint["date"], tz=timezone.utc
+                ).strftime("%Y-%m-%d"),
```
lithium323, line 59: from op_coreutils.time import dt_fromepoch

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +84,102 @@ def process_breakdown_stables(data):
...
+        A dictionary containing metadata and breakdown DataFrames.
     """
     session = new_session()
     summary = get_data(session, SUMMARY_ENDPOINT)
 
-    urls = {}
-    for stablecoin in summary["peggedAssets"]:
-        stablecoin_id = stablecoin["id"]
-        urls[stablecoin_id] = BREAKDOWN_ENDPOINT.format(id=stablecoin_id)
+    all_stablecoin_ids = [asset["id"] for asset in summary.get("peggedAssets", [])]
```
lithium323, line 120: I think we must not have a fallback for `peggedAssets` like this:
```
summary.get("peggedAssets", [])
```
We want the script to fail if the `peggedAssets` key is not present.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +84,102 @@ def process_breakdown_stables(data):
...
+        if not stablecoin_ids:
+            log.error("No valid stablecoin IDs provided.")
+            return {"metadata": pl.DataFrame(), "breakdown": pl.DataFrame()}
+    else:
+        stablecoin_ids = all_stablecoin_ids
+
+    urls = {
+        stablecoin_id: BREAKDOWN_ENDPOINT.format(id=stablecoin_id)
+        for stablecoin_id in stablecoin_ids
+    }
```
lithium323, line 133:
```suggestion
    urls = {}
    for stablecoin in summary["peggedAssets"]:
        stablecoin_id = stablecoin["id"]
        if stablecoin_id in (stablecoin_ids or []):
            urls[stablecoin_id] = BREAKDOWN_ENDPOINT.format(id=stablecoin_id)
            
    if not urls:
        raise ValueError("no valid stablecoin IDs provided.")
```

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +84,102 @@ def process_breakdown_stables(data):
...
-
-    # Write breakdown to BQ. Only the most recent 7 days are inserted.
-    # Data in BQ older than 7 days is considered immutable.
-    overwrite_partitions_dynamic(
-        df=most_recent_dates(breakdown_df, n_dates=7),
-        dataset=BQ_DATASET,
-        table_name=f"{BREAKDOWN_TABLE}_history",
-    )
+
+    if not metadata_df.is_empty():
```
lithium323, line 162: we should never get empty dataframes should we?  If we do then should we raise an exception instead of silently not writing to BQ?

lithium323 (changes requested):
leaving another round of comments, still haven't gone through the tests part yet

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
-    - Fetch the stablecoins summary endpoint.
-    - For each stablecoin, fetch the detailed breakdown data.
-    - Write all results to BigQuery.
+    Returns:
+        A dictionary containing metadata and breakdown DataFrames.
     """
     session = new_session()
     summary = get_data(session, SUMMARY_ENDPOINT)
 
+    if "peggedAssets" not in summary:
```
lithium323, line 127: In the next line we do have `summary["peggedAssets"]` which will raise the KeyError if `peggedAssets` is not there. In the exception traceback we can see the line of code where the exception came from and that should give us the right context to understand the error.  I think it's ok to rely on system exceptions for the most data shape checks.  Or else we can end up reimplementing a lot of checks.   For example, what would happen if `summary["peggedAssets"]` is not an iterable.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
+        raise KeyError("The 'peggedAssets' key is missing from the summary data.")
+
     urls = {}
     for stablecoin in summary["peggedAssets"]:
         stablecoin_id = stablecoin["id"]
-        urls[stablecoin_id] = BREAKDOWN_ENDPOINT.format(id=stablecoin_id)
+        if stablecoin_ids is None or stablecoin_id in stablecoin_ids:
```
lithium323, line 133: I think I may have suggested this, but now that I read it again maybe the following is even more readable:
```
if stablecoin_id in (stablecoind_ids or []):
```

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
+        lambda x: get_data(session, x), urls, max_workers=4
+    )
 
-    breakdown_dfs = []
-    metadata_rows = []
-    for data in stablecoin_data.values():
-        breakdown_df, metadata = process_breakdown_stables(data)
-        breakdown_dfs.append(breakdown_df)
-        metadata_rows.append(metadata)
+    breakdown_dfs: List[pl.DataFrame] = []
```
lithium323, line 142: nice, love having types! 

Since a recent python version, can't remember which one, you don't need to import the special capitalized `List` or `Dict`.   You can use:
```
breakdown_dfs: list[pl.DataFrame] = []
metadata_rows: list[dict[str, str]] = []
```

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
+    metadata_rows: List[Dict[str, Optional[str]]] = []
 
-    breakdown_df = pl.concat(breakdown_dfs, how="diagonal_relaxed")
-    metadata_df = pl.DataFrame(metadata_rows, infer_schema_length=len(metadata_rows))
+    for data in stablecoin_data.values():
+        breakdown_df, metadata = process_breakdown_stables(data, days=days)
+        if not breakdown_df.is_empty():
```
lithium323, line 147: I may be missing something, but what would be the cases when `breakdown_df` is empty?   Or when metadata would be a falsy value?  

Should we be raising exceptions from within `process_breakdown_stables` in those cases instead?

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
+            breakdown_dfs.append(breakdown_df)
+            if metadata:
+                metadata_rows.append(metadata)
+
+    breakdown_df = (
+        pl.concat(breakdown_dfs, how="diagonal_relaxed")
+        if breakdown_dfs
+        else pl.DataFrame()
+    )
+    metadata_df = (
```
lithium323, line 157: this is addressing my own code that I wrote before. Maybe it would be good to have `process_breakdown_stables` return a metadata dataframe instead of a metadata list of dicts. That way it would be more uniform and we could use pl.concat() jus like we do for the breakdown df.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
+    for data in stablecoin_data.values():
+        breakdown_df, metadata = process_breakdown_stables(data, days=days)
+        if not breakdown_df.is_empty():
+            breakdown_dfs.append(breakdown_df)
+            if metadata:
+                metadata_rows.append(metadata)
+
+    breakdown_df = (
+        pl.concat(breakdown_dfs, how="diagonal_relaxed")
+        if breakdown_dfs
```
lithium323, line 154: I don't thing there would be an acceptable case where we don't get any `breakdown_dfs`. So we should have an exception raised somewhere rather than the special case handling when concatenating.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
 
-    # Write breakdown to BQ. Only the most recent 7 days are inserted.
-    # Data in BQ older than 7 days is considered immutable.
-    overwrite_partitions_dynamic(
-        df=most_recent_dates(breakdown_df, n_dates=7),
+    if metadata_df.is_empty():
+        raise ValueError(
+            "metadata_df is empty. Expected non-empty data to write to BigQuery."
+        )
+    overwrite_unpartitioned_table(metadata_df, BQ_DATASET, f"{METADATA_TABLE}_latest")
```
lithium323, line 169: I had this call to `overwrite_unpartitioned_table` the first time I implemented this. But I meant to ask both you and @chucktraverse  if you think this table is really needed. Maybe we don't need to save this `_latest` version of the data.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -88,60 +83,111 @@ def process_breakdown_stables(data):
...
-        table_name=f"{BREAKDOWN_TABLE}_history",
+        table_name=f"{METADATA_TABLE}_history",
     )
 
+    if breakdown_df.is_empty():
+        raise ValueError(
+            "breakdown_df is empty. Expected non-empty data to write to BigQuery."
+        )
+
+    dates = breakdown_df["dt"].unique().to_list()
```
lithium323, line 182: I think that the functionality of splitting the data frame into `dt` parts and making sure each partition is upserted accordingly should be the responsibility of the `upsert_partitioned_table` function.  This is almost certainly going to come up again each time we want to use `upsert_partitioned_table`.

lithium323 (changes requested):

src/op_analytics/cli/subcommands/pulls/defillama.py
lithium323: hmm, re-reading my comment my logic is incorrect it wouldn't do the same as what we want.  So ok to keep it as is

dioptx: I am a bit late in answering this but yes I noticed through building the tests that the two versions are not equivalent. Keeping this as is for now 🙏

src/op_analytics/cli/subcommands/pulls/defillama.py
dioptx: Yeah it was in the spirit of personal preference, something like "SQL native commands like 'FROM' should be capitalized" to not confuse them with the rest. 

Will revert

src/op_analytics/cli/subcommands/pulls/defillama.py
dioptx: Replaced with exceptions in `process_breakdown_stables`

tests/op_analytics/cli/pulls/defillama/test_defillama.py
```diff
@@ -1,149 +1,244 @@
...
+}
+
+another_sample_breakdown_data = {
+    # Similar structure with required fields
+    "pegType": "peggedUSD",
+    "chainBalances": {
+        "Binance Smart Chain": {
+            "tokens": [
+                {
+                    "date": current_timestamp,
```
lithium323, line 70: I would advise against more complex test data like this. Instead maybe the function you are testing should accept cutoff_date directly rather than accepting the timedelta from the present.  That way you can keep your test data simpler. Ideally test data would be straight up copied from a real API response without any modifications.

src/op_analytics/cli/subcommands/pulls/defillama.py
```diff
@@ -22,50 +24,40 @@
...
-def get_data(session, url):
-    start = time.time()
-    resp = session.request(
-        method="GET",
-        url=url,
-        headers={"Content-Type": "application/json"},
-    ).json()
-    log.info(f"Fetched from {url}: {time.time() - start:.2f} seconds")
-    return resp
+def process_breakdown_stables(
```
lithium323, line 27: now that I have been going over this data pull more carefully I see that the process_breakdown_stables function is doing two separate things, so it should be broken apart into two functions.  One is getting the metadata, the other one is getting the balances.   

At the time when we first coded this up we put it toghether somewhat quickly and were not very careful about things like this. Also the names we used could have been a lot better. For example, `balances` would be more indicative of what is going and therefore a better name than `breakdown`

lithium323 (approved):

PR #979, opened 2024-11-06
All my modifications after very thorough review of Defillama stablecoins data pull

lithium323:
Some notes:

- Instead of filtering by stablecoin ids suggest that we filter by symbol since it is more human friendly.
- Finished plugging the filter to the CLI command so it can be used when calling the CLI.
- Modified the biguqery upssert partitioned table function so it can handle dataframes that have multiple dt values.
- Also added a flag to create the table if upserting for the first time.
- Revisit the bigquery table names and only make two calls to write to bigquery. We were saving a history table that does not seem necessary.
- Pull apart more functions in defillama.py to make it easier to parse. 
- Renamed various things, we used the name `breakout` but that doesn't really tell us much. Using `balances` which I think is better but open to suggestions. 
-  Made the `n_days` parameter a constant instead. Wasn't seeing the need to expose it as a parameter.

### DIFF
--- a/packages/op-coreutils/src/op_coreutils/bigquery/write.py
+++ b/packages/op-coreutils/src/op_coreutils/bigquery/write.py
@@ -51,6 +51,12 @@ class OPLabsBigQueryError(Exception):
     pass
 
 
+class OPLabsUpsertTableNotExists(Exception):
+    """Exception raised when an upserted table does not exist yet."""
+
+    pass
+
+
 def overwrite_unpartitioned_table(df: pl.DataFrame, dataset: str, table_name: str):
     """Overwrite an unpartitioned BigQuery table with the given DataFrame.
 
@@ -63,9 +69,6 @@ def overwrite_unpartitioned_table(df: pl.DataFrame, dataset: str, table_name: st
         OPLabsBigQueryError: If the table name does not end with '_staging' or '_latest'.
     """
     destination = f"{dataset}.{table_name}"
-
-    if not any([table_name.endswith("_staging"), table_name.endswith("_latest")]):
-        raise OPLabsBigQueryError(f"cannot overwrite data at {destination}")
 
     _write_df_to_bq(
         df,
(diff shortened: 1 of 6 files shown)


packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -63,9 +70,6 @@ def overwrite_unpartitioned_table(df: pl.DataFrame, dataset: str, table_name: st
     """
     destination = f"{dataset}.{table_name}"
 
-    if not any([table_name.endswith("_staging"), table_name.endswith("_latest")]):
```
lithium323, line 66: this was really more of a nuisance

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -153,25 +157,10 @@ def overwrite_partitions_dynamic(
         expiration_days (int, optional): If provided sets the expiration time of the table.
     """
 
-    # Ensure "dt" is a DateTime
```
lithium323, line 156: The logic we were applying to the dataframe here, checking it has `dt` and breaking it into `dt` parts is made into a function so we can reuse it in the `upsert_partitioned_dataset` function

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -212,6 +201,40 @@ def overwrite_partition_static(
...
+    if df["dt"].dtype == pl.String:
+        df = df.with_columns(dt=pl.col("dt").str.strptime(pl.Datetime, "%Y-%m-%d"))
+
+    partitions = df["dt"].unique().sort().to_list()
+
+    log.info(
+        f"Found {len(partitions)} partitions in dataframe [{partitions[0]} ... {partitions[-1]}]"
+    )
+
+    if len(partitions) > 10:
```
lithium323, line 227: maybe in the future we can make this a parameter instead of limiting it to 10 partitions. For now I don't see cases where we might need to update that many partitions at once. It would defeat the purpose of incremental processing.

dioptx (approved):

packages/op-coreutils/src/op_coreutils/bigquery/write.py
dioptx: Agreed, adding a todo: comment here so that we keep track of this.

PR #994, opened 2024-11-09
Update github_analytics.py with new repos

baokiddy:
**Description**

Adding additional OP public repos to the list so we have all traffic analytics for them accounted for.

**Tests**

No tests as the script is functionally the same

**Additional context**

Need to work on PR and commit analytic tables for these repos as well 
cc @lithium323

### DIFF
--- a/src/op_analytics/cli/subcommands/pulls/github_analytics.py
+++ b/src/op_analytics/cli/subcommands/pulls/github_analytics.py
@@ -20,7 +20,7 @@ log = structlog.get_logger()
 REPOS_BASE_URL = "https://api.github.com/repos/ethereum-optimism"
 
 # Repos to track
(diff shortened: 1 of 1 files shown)


lithium323:
> **Description**
> 
> Adding additional OP public repos to the list so we have all traffic analytics for them accounted for.
> 
> **Tests**
> 
> No tests as the script is functionally the same
> 
> **Additional context**
> 
> Need to work on PR and commit analytic tables for these repos as well cc @lithium323

Hi @baokiddy  the PR looks good.  You should not need to create new tables for the repos, all the data can go in the same table in BQ. There is a `repo_name` column there that you can use during analysis to tell them apart.

lithium323 (approved):

PR #1010, opened 2024-11-14
[Do Not Merge] Defillama chain tvl feedback request

chucktraverse:
Still working on tests, just making a PR for review purposes


packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -209,7 +209,7 @@ class DatePart:
     date_suffix: str
 
 
-def breakout_partitioned_df(df: pl.DataFrame) -> Generator[DatePart, None, None]:
+def breakout_partitioned_df(df: pl.DataFrame, run_check=True) -> Generator[DatePart, None, None]:
```
lithium323, line 212: wondering what your thoughts are behind the flag.  I'll share mine. 

I guess it would be good to first clarify when we should and shouldn't use partitioned tables and that might help us converge. When I think of a partitioned table I see a table for which new data will only modify at most a few of the most recent partitions (close to the current date).  Older partitions never change and so should remain untouched. 

So the 10 partitions check is there to remind people that even if the data you are obtaining has information for much older dates you should not be writing all of it all over again, only updating the most recent partitions.

chucktraverse:
@mslipper @chuxinh is the metadata portion of this data pull even useful? Can it feed into other downstream processes or is all of this data tracked in a google sheet that get's referenced in other data pulls?

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -287,7 +288,7 @@ def upsert_unpartitioned_table(
     Raises:
         ValueError: If the DataFrame is empty or if unique_keys are not in the DataFrame.
     """
-    if "dt" in df.columns:
+    if "dt" in df.columns and not allow_unpartitioned_dt:
```
lithium323, line 291: I think we should stick to this convention in our data warehouse. If you see a table with a `dt` column it should imply that it is partitioned by `dt`.  

When unpartitioned data needs to have a date column we could name the column `date`.

src/op_analytics/cli/subcommands/pulls/app.py
```diff
@@ -4,7 +4,10 @@
 from op_coreutils.logger import structlog
 
 from .agora import agora_pull
-from .defillama import pull_stablecoins as dfl_pull_stablecoins
+from .defillama.defillama_stablecoins import pull_stablecoins as dfl_pull_stablecoins
```
lithium323, line 7: the import has a redundant ring to it.  It might be better if it was `from .defillama.stablecoins`  and `from .defillama.historical_chain_tvl` 

Also on the `historical` qualifier.  I suggest we drop it from the name.  Something like 
```
from .defillama.tvl_chains
```
I wasn't sure what the `historical` name was implying.  

Also the reason I suggest `tvl_chains` is that then the modules would be sorted alphabetically more nicely if you later on add something like `tvl_apps`.

src/op_analytics/cli/subcommands/pulls/defillama/defillama_historical_chain_tvl.py
```diff
@@ -0,0 +1,201 @@
...
+
+CHAINS_METADATA_ENDPOINT = "https://api.llama.fi/config"
+CHAINS_ENDPOINT = "https://api.llama.fi/v2/chains"
+CHAINS_TVL_ENDPOINT = "https://api.llama.fi/v2/historicalChainTvl/{slug}"
+
+BQ_DATASET = "temp"
+
+CHAIN_METADATA_TABLE = "defillama_chains_metadata"
+HISTORICAL_CHAIN_TVL_TABLE = "defillama_daily_historical_chain_tvl"
+TVL_TABLE_LAST_N_DAYS = 100_000  # upsert only the last 7 days of tvl records fetched from the api
```
lithium323, line 22: here is where I think we need to agree on what the incremental updates approach will be.  Do we need to re-upsert the last 100k days every time?

src/op_analytics/cli/subcommands/pulls/defillama/__init__.py
lithium323, line 1: I was expecting the existing `defillama.py` file to show up as deleted in the PR.

lithium323 (commented):
leaving an initial round of comments.

src/op_analytics/cli/subcommands/pulls/app.py
chucktraverse: it's a reference to the name of the end point. Defillama has chains, tvl, and historicalchainstvl end points. I wanted to keep it as similar to the end point name as possible so it's super clear what exactly we're pulling

open to changing though and decoupling from the naming conventions within defillama

src/op_analytics/cli/subcommands/pulls/defillama/__init__.py
chucktraverse: You're right. I had kept it around as a reference and forgot to delete. It'll come in deleted in the next push

src/op_analytics/cli/subcommands/pulls/defillama/defillama_historical_chain_tvl.py
chucktraverse: This is part of the "Do Not Merge" part as I'm still using this for testing purposes. I'll change it back to 7 days and change BQ_DATASET to "api_uploads".  I also need to flip the create new table flags to False

packages/op-coreutils/src/op_coreutils/bigquery/write.py
chucktraverse: Agreed, and will change accordingly

src/op_analytics/cli/subcommands/pulls/app.py
lithium323: ah that makes perfect sense then let's honor the API names

src/op_analytics/cli/subcommands/pulls/defillama/defillama_historical_chain_tvl.py
lithium323: sounds good! my recommendation while you are testing would be to comment out the upsert and replace with an overwrite call altogether.  This includes commenting out the part where the last N days are selected from the data.  

Otherwise the upsert utility will get in your way if you are trying to update too many partitions at once.

packages/op-coreutils/src/op_coreutils/bigquery/write.py
```diff
@@ -88,17 +88,17 @@ def most_recent_dates(df: pl.DataFrame, n_dates: int) -> pl.DataFrame:
 
     Assumes the input dataframe has a "dt" column.
     """
+    date_schema = "date" if "date" in df.columns else "dt"
```
lithium323, line 91: I went ahead and added support for a different column on the most_recent_dates function. That way it will be more ergonomic to use for the dataframe with a "date" column.  https://github.com/ethereum-optimism/op-analytics/pull/1014

packages/op-coreutils/src/op_coreutils/bigquery/write.py
lithium323, line 1: this file looks like a `packages/...` path that should be removed (maybe a leftover from before the packages were consolidated)

chucktraverse:
Closing for an updated version with both protocol and historical chain tvl data pulls

PR #1056, opened 2024-11-22
Agora implement fetch functions

dioptx:
### Objective:
Introduce a new module for fetching data from the Agora API, including endpoints for delegates, proposals, and votes, with pagination and checkpoint management.

### Key Changes:

- Added fetch.py for API interactions with pagination support.
- Implemented CheckpointManager for managing pagination checkpoints.
- Updated get_data in
- request.py to support query parameters.
- Added unit tests for fetch functions in test_fetch.py.

### Rationale & Future Work:
This PR focuses on data fetching logic. Subsequent PRs will address data processing and table updates, maintaining smaller, focused code sets.


src/op_analytics/coreutils/request.py
```diff
@@ -55,12 +56,12 @@ def get_data(
         with attempt:
             if attempt.num > 1:
                 log.warning(f"retrying {url}")
-            return _get_data(session, url, headers)
+            return _get_data(session, url, headers, params)
 
 
-def _get_data(session: requests.Session, url: str, headers: dict[str, str]):
+def _get_data(session: requests.Session, url: str, headers: dict[str, str], params: dict[str, str]):
```
lithium323, line 62: I think the type hint for params should be `dict[str, Any]`. In case param values are not strings (for example ints or lists).

src/op_analytics/cli/subcommands/pulls/agora/fetch.py
```diff
@@ -0,0 +1,116 @@
...
+
+    while True:
+        query_params = {"limit": limit, "offset": offset}
+        if params:
+            query_params.update(params)
+
+        response = get_data(session, url, headers=HEADERS, params=query_params)
+
+        # Check if the response is a list or a dictionary
+        if isinstance(response, list):
```
lithium323, line 47: is the toggle here to support endpoints that behave differently, or is it expected that the same endpoint switches the response payload from dict to list?  

if it is to support endpoints that behave differently I would suggest having a separate pagination helper.

src/op_analytics/cli/subcommands/pulls/agora/fetch.py
```diff
@@ -0,0 +1,116 @@
...
+            batch = response.get("data", [])
+
+        data.extend(batch)
+
+        # If the response is a list, we assume there's no pagination metadata
+        if isinstance(response, list) or not response.get("metadata", {}).get("has_next", False):
+            break
+
+        # Update offset for the next batch
+        offset = response.get("metadata", {}).get("next_offset", offset + limit)
```
lithium323, line 59: I'm curious about the fallbacks. If we don't have metadata, should we fail instead? I'm assuming that not having "metadata" would be the API behaving in an unexpected way, so working around it could lead to data integrity problems for the ingested data.

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,204 @@
...
+    )
+
+    return AgoraDelegates(
+        delegates_with_voting_power_df=delegates_with_voting_power,
+        delegates_without_voting_power_df=delegates_without_voting_power,
+    )
+
+
+# Todo: move to app.py
+def pull_delegate_data():
```
thbialek, line 188: I know we are pre-filtering for delegates with voting power > 0 for performance reasons, but just to confirm — we are still pulling data for delegates with voting power = 0, right?

I think it definitely makes sense to trim the delegate set down to those with non-zero voting power for votes, but think we'd still want to be able to track voting power balances over time for delegates whose voting power is 0 today (e.g. if a delegate had non-zero vp and voted on proposals last season, but then lost all their vp this season for whatever reason).

thbialek (approved):

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
dioptx: Correct! We will be saving both cases but pulling events for delegates that have a non zero balance on a given date. 
An edge case to this, is if a delegate:
1. Gets voting power
2. Performs an event
3. Looses all the voting power before recorded in the delegates table

We would need to pull events for all delegates on a larger cadence to account for this, because Agora's API does not provide a way for us to pull recent votes and it should be time consuming to pull all the delegate information every day.

thbialek: Understood, thanks for clarifying! In that case, only extremely rare edge cases would be affected (i.e. intra-day swings from VP > 0 to VP = 0 coupled with a vote event on the same day), which we will eventually be able to fill in once we found a workaround solution for the limitations of the Agora API.

src/op_analytics/cli/subcommands/pulls/agora/data_access.py
```diff
@@ -0,0 +1,237 @@
...
+
+log = structlog.get_logger()
+
+MARKERS_TABLE = "daily_data_markers"
+
+
+class Agora(str, Enum):
+    """Supported agora datasets."""
+
+    DELEGATES = "delegates_v1"
```
lithium323, line 34: wanted to better understand the reason to split the data into with/without voting power. Would it be possible to keep voting power as a column and have a single dataset?

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,170 @@
...
+@dataclass
+class SimplePaginator:
+    url: str
+    limit: int = 50
+
+    @retry(
+        stop=stop_after_attempt(3),
+        wait=wait_exponential(multiplier=1, min=10, max=40),
+        retry=retry_if_exception_type(
+            (requests.exceptions.Timeout, requests.exceptions.ConnectionError)
```
lithium323, line 49: The requests session that we use already mounts a Retry adapter. We can set the number of connect error retries there. Right now we don't specify it so it deafults to 0.

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,170 @@
...
+    parse_isoformat_with_z,
+)
+from op_analytics.coreutils.logger import structlog
+from op_analytics.coreutils.request import get_data
+from op_analytics.coreutils.threads import run_concurrently
+
+# from op_analytics.coreutils.time import parse_isoformat_with_tz
+from typing import Any
+import requests
+from tenacity import retry, stop_after_attempt, wait_exponential, retry_if_exception_type
```
lithium323, line 15: lets use "stamina" instead of "tenacity".  We are already using stamina on other parts so it would be good to make that uniform.   

That said I think we should try to rely only on the two types of retries that are supported on request.py.  We have a Retry adapter and then also a retrier that retries on all exceptions just in case (we use this for defillama where the kind of error we may get is not always clear).

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,213 @@
...
+        return list(itertools.chain.from_iterable(data.values()))
+
+
+def pull_delegates():
+    """Pull and write agora delegates data."""
+    p = Paginator(url=f"{BASE_URL}/delegates", params={"sort": "voting_power"}, max_workers=32)
+    delegate_data = p.concurrent_fetch()
+
+    # Clean and transform to Pandas
+    df = pd.json_normalize(delegate_data, sep="_")
```
lithium323, line 148: Here it would be nice to do something that helps us understand what the schema is that we are getting. Instead of going to pandas, could we normalize the data ourselves in a python loop? 

I think a pure python loop instead of pandas might be easier to grasp when reading the code.

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,213 @@
...
+    for col in ["voting_power_total", "voting_power_direct", "voting_power_advanced"]:
+        df[col] = pd.to_numeric(df[col], errors="coerce")
+
+    df["dt"] = (
+        now_dt()
+    )  # Todo: This is to partition by pull date. Figure out if this is the correct way to do this.
+    # Convert to polars
+    df = pl.from_pandas(df)
+
+    # Split to delegates that have voting power and those that don't
```
lithium323, line 169: not sure about why splitting into two tables. Could we get away with a single table and having readers do the filtering?

src/op_analytics/cli/subcommands/pulls/agora/data_access.py
dioptx: I had this thought initially because of the assymetric nature of the data (very few with voting power vs the rest), and how we would need to pull the entire table and filter for voting power each time we want to do an analysis.

On a second thought we can leave this out, since it now adds complexity that I'm not 100% is needed.

tests/op_analytics/cli/pulls/agora/test_fetch.py
```diff
@@ -0,0 +1,61 @@
+from unittest.mock import patch, MagicMock
+import requests
+from op_analytics.cli.subcommands.pulls.agora.fetch import (
+    fetch_proposals,
+)
+
+
+# Test for pagination
+@patch("op_analytics.cli.subcommands.pulls.agora.fetch.get_data")
```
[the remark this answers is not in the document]
dioptx: You are right, these are left overs and will be replaced with new ones, removing.

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,213 @@
...
+    write,
+    camelcase_to_snakecase,
+)
+
+log = structlog.get_logger()
+
+BASE_URL = "https://vote.optimism.io/api/v1"
+DELEGATES_ENDPOINT = f"{BASE_URL}/delegates"
+API_KEY = env_get("AGORA_API_TOKEN")
+session = requests.Session()  # Create a session object
```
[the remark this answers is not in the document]
dioptx: The pains of translating jupyter notebook code, fixing

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
dioptx: Reinvented the wheel here, removing all extra logic and using the retry logic provided by requests.py

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
dioptx: Fixed as mentioned above

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
dioptx: Used the proper functionality from request.py for this

src/op_analytics/coreutils/request.py
```diff
@@ -4,7 +4,7 @@
 import requests
```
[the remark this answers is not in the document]
dioptx: Good idea, here's a tiny PR for this: #1103

src/op_analytics/cli/subcommands/pulls/agora/data_access.py
```diff
@@ -0,0 +1,235 @@
...
+            location=DataLocation.GCS,
+        )
+
+    def dump_local_copy(
+        self,
+        min_date: str | None = None,
+        max_date: str | None = None,
+        date_range_spec: str | None = None,
+    ):
+        """Dump a copy of the data to a single local parquet file."""
```
lithium323, line 57: for all of the methods in the dataaccess module, could you take a look at defillama and use the functions that were lifted over to coreutils for common use? 

Also could you rename data_access.py to dataaccess.py?  that's just for conforming to the name we use in defillama.  For modules that are meant to be imported frequently I try to not use underscores.

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,161 @@
...
+import requests.exceptions
+import pandas as pd
+import os
+
+log = structlog.get_logger()
+
+BASE_URL = "https://vote.optimism.io/api/v1"
+DELEGATES_ENDPOINT = f"{BASE_URL}/delegates"
+API_KEY = os.environ["AGORA_API_KEY"]
+session = requests.Session()  # Create a session object
```
lithium323, line 23: this session object also needs to be scoped inside a class or function

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,161 @@
...
+            else:
+                log.error(
+                    f"HTTP Error for address: {address}, URL: {e.response.url}, Status Code: {e.response.status_code}"
+                )
+            failed_addresses.append(address)
+            return []
+
+    all_delegate_data = []
+
+    for i in range(0, len(delegates), batch_size):
```
lithium323, line 95: Using batches when we already parallelized with threads seems unnecessary.  

Also, instead of a second pass of retrying, my recommendation would be to decorate `_fetch_with_fallback` with `@stamina.retry` and that way you can have the threads themselves take care of retries. 

If you want to tolerate some failures, because retries are not enough, take a look at `run_concurrently_store_failures` that seems better than catching exceptions like we are doing in `_fetch_with_fallback`.

src/op_analytics/cli/subcommands/pulls/agora/data_access.py
dioptx: I observe that we use Pandas and Polars at the same time so far in both Defillama and Agora. 
If there isn't a reason for Polars to exist, should I port dataaccess.py to using only Pandas?
(re. dataaccess, I thought the missing "_" was a typo) 😁

lithium323: I tend to use polars more, but if there's functionality you'd rather implement in pandas then I'm all for it as well.  Right now I don't think the defillama pulls use pandas. Maybe Chuck uses it for analysis of the pulled data?

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,143 @@
...
+
+        return all_data
+
+
+def fetch_event_data(delegates: list, endpoint: str, workers: int) -> pd.DataFrame:
+    def fetch_delegate_data(address: str, endpoint: str) -> list:
+        paginator = SimplePaginator(url=f"{BASE_URL}/delegates/{address}/{endpoint}", limit=50)
+        try:
+            return paginator.fetch_all()
+        except requests.exceptions.HTTPError as e:
```
lithium323, line 80: If we catch the exception then the error will not be detected as a failure by `run_concurrently_store_failures`, I think we have to let exceptions bubble up.

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,143 @@
...
+    all_delegate_data = run_results.results.values()
+    failed_addresses = run_results.failures.keys()
+
+    if len(failed_addresses) > 0:
+        log.error(f"Failed to fetch data for {len(failed_addresses)} addresses")
+
+    return all_delegate_data
+
+
+def _flatten_data(data):
```
lithium323, line 102: what do you think about being more explicit about the keys that you pick from values? Something like:
```
def _flatten_data(data):
    rows = []
    for record in data:
        for address, values in record.items():
            rows.append(
                {
                    "address": address,
                    "requiredX": values["requiredX"],
                    "optionalY": values.get("optionalY"),
                }
            )
    return rows 
```

I think that can help us be more aware of any possible changes in the api.

lithium323: Also if you are more explicit you can include all of your transformations in the flattening loop instead of using pandas on the overall dataframe.

src/op_analytics/cli/subcommands/pulls/agora/delegate_events.py
```diff
@@ -0,0 +1,143 @@
...
+
+def fetch_proposals():
+    paginator = SimplePaginator(url=f"{BASE_URL}/proposals", limit=50)
+    proposals = paginator.fetch_all()
+    df = pd.DataFrame(proposals)
+    df.columns = df.columns.map(_camelcase_to_snakecase)
+    df.set_index("start_time", inplace=True)
+    df.index = df.index.map(parse_isoformat_with_z)
+
+    write(dataset=Agora.PROPOSALS, dataframe=df, sort_by=["start_time"])
```
lithium323, line 143: the dataframe expected by the write api will be a polars dataframe (here it looks like we pass a pandas dataframe).  

My advise would be the same as above in terms of being more explicit about which columns you are picking from the result. I think that' important to maintain the code longer term, because by reading the code (without running it) you get an idea of the exact columns and types that are being used.

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,232 @@
...
+                    "statement_signature": statement.get("signature", ""),
+                    "statement_created_at": statement.get("created_at", ""),
+                    "statement_updated_at": statement.get("updated_at", ""),
+                    "statement_text": statement.get("payload", {}).get("delegateStatement", ""),
+                    "twitter": statement.get("payload", {}).get("twitter", ""),
+                    "discord": statement.get("payload", {}).get("discord", ""),
+                }
+            )
+
+        normalized_data.append(flattened)
```
lithium323, line 173: +1 on this approach! I like being able to get an idea of what the data will look like as I read the code.

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,232 @@
...
+    for item in delegate_data:
+        # Extract fields from the item dictionary
+        address = item.get("address", "")
+        voting_power = item.get("votingPower", {})
+        citizen = item.get("citizen", False)
+        statement = item.get("statement", {})
+
+        # Make sure voting_power is a dictionary
+        if not isinstance(voting_power, dict):
+            continue
```
lithium323, line 151: should we raise an exception here, or maybe just not continue and let python raise an exception on the next block where we try to use it as a dictionary?

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
lithium323: in the loop, every value has a fallback. Are there some values that are considered required so we would rather have an exception than a fallback?

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,232 @@
...
+    if df.shape[0] != len(df["address"].drop_duplicates()):
+        df = df.drop_duplicates()
+        log.warning(
+            f"Found duplicates in delegates data. Dropped {df.shape[0] - len(df['address'].drop_duplicates())} rows."
+        )
+
+    # Todo: This is to partition by pull date. Figure out if this is the correct way to do this.
+    df["dt"] = now_dt()
+
+    # Convert to polars
```
lithium323, line 201: Maybe we should stay in polars world from the start?  Given that it seems we are not using any functionality that is very specific to pandas.

src/op_analytics/cli/subcommands/pulls/agora/delegates.py
```diff
@@ -0,0 +1,207 @@
...
+    # Add a dt column for partitioning
+    df = df.with_columns(pl.lit(now_dt()).alias("dt"))
+
+    # Write to GCS using Agora's write
+    Agora.DELEGATES.write(dataframe=df, sort_by=["voting_power_total"])
+
+    return AgoraDelegates(delegates_df=df)
+
+
+# Todo: move to app.py
```
lithium323, line 185: Let's go ahead and move this to app.py. I was going to test it locally and realized I could not yet access the code from the CLI.

PR #1057, opened 2024-11-22
Add token value

chucktraverse:
**Description**

**Tests**

**Additional context**

**Metadata**


chucktraverse:
TO DO
Update tests to generate time epochs based on the current time. As of now they are hard coded and will move outside of the 7 day window, causing the code to skip those dummy records and fail the tests

PR #1123, opened 2024-12-09
Extend functionality for threads.py

dioptx:
Extends the functionality of theads.py to include static function_args to the functions that are parallelized.

### DIFF
--- a/src/op_analytics/coreutils/threads.py
+++ b/src/op_analytics/coreutils/threads.py
@@ -12,6 +12,7 @@ def run_concurrently(
     function: Callable,
     targets: dict[str, Any] | list[Any],
     max_workers: int | None = None,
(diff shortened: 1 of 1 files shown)


lithium323 (changes requested):
there can be a lot of gotchas when implementing function args. A tuple is tricky because it has an implied ordering, if you want to support args you almost also have to have support for kwargs.  Rather than making our implementation here more complicated I think it is reasonable to expect that callers provide a single argument callable.

My recommendation would be to add something on the docstring of run_concurrently:
```
Sometimes you have function that accepts arguments in addition to the function target
that you need to run concurrently. For those cases we recommend that you create a 
single arg function wrapper at the call site which you can then pass to run_concurrently.
```

PR #1126, opened 2024-12-09
Add new functionality to parse isoformats to time.py

dioptx:
We are ingesting datetime strings of this manner: "2023-07-09T13:47:17.000Z" from the Agora Api and currently they cannot be processed via any function available in time.py.

Specifically

- The date_fromstr function expects a date string in YYYY-MM-DD format, not a full ISO timestamp.
- Other functions like datetime_fromepoch or date_toepoch work with epoch values and not ISO 8601 strings.

This small PR introduces a way to achieve that.

### DIFF
--- a/src/op_analytics/coreutils/time.py
+++ b/src/op_analytics/coreutils/time.py
@@ -64,3 +64,10 @@ def surrounding_dates(dateval: date) -> list[date]:
 def epoch_is_date(epoch: int) -> bool:
     """Return true if seconds since epoch is at exactly midnight."""
     return epoch % 86400 == 0
+
+
+def parse_isoformat_with_z(iso_string):
+    """Parse ISO format strings with 'Z' to datetime objects."""
+    if iso_string.endswith("Z"):
+        iso_string = iso_string[:-1] + "+00:00"
+    return datetime.fromisoformat(iso_string)


src/op_analytics/coreutils/time.py
```diff
@@ -64,3 +64,10 @@ def surrounding_dates(dateval: date) -> list[date]:
 def epoch_is_date(epoch: int) -> bool:
     """Return true if seconds since epoch is at exactly midnight."""
     return epoch % 86400 == 0
+
+
+def parse_isoformat_with_z(iso_string):
```
lithium323, line 69: nice, this is great to have! 

I noticed that the Z suffix does work out of the box:
```
In [1]: from datetime import datetime

In [2]: datetime.fromisoformat("2023-07-09T13:47:17.000Z")
Out[2]: datetime.datetime(2023, 7, 9, 13, 47, 17, tzinfo=datetime.timezone.utc)
```
but maybe I'm missing something about an edge case? If so we should comment why we handle the Z as a special case. 

Also it's important to note that this method is returning a datetime with timezone. Using fromisoformat can be tricky since in some cases it returns tzinfo and in some cases it does not. 
```
In [5]: datetime.fromisoformat("2023-07-09T13:47:17.000").tzinfo

In [6]: datetime.fromisoformat("2023-07-09T13:47:17.000Z").tzinfo
Out[6]: datetime.timezone.utc
```

Functions in the time module always return naive datetimes (all callers can safely assume the object is UTC and should not ever have to worry about handling timezones), so wondering if here we would prefer to ensure that we return a naive datetime object always.

lithium323: hey! I see the function now returns a datetime with timezone.  It should return a naive datetime (tzinfo == None)

src/op_analytics/coreutils/time.py
```diff
@@ -64,3 +64,13 @@ def surrounding_dates(dateval: date) -> list[date]:
 def epoch_is_date(epoch: int) -> bool:
     """Return true if seconds since epoch is at exactly midnight."""
     return epoch % 86400 == 0
+
+
+def parse_isoformat(iso_string):
+    """Add 'Z' suffix if not present and parse ISO format strings to datetime objects."""
+    if iso_string.endswith("Z"):
+        iso_string = iso_string[:-1] + "+00:00"  # Replace 'Z' with '+00:00' for parsing
```
lithium323, line 72: still not sure why the 'Z' suffix is handled as a special case, it seems to work well with fromisoformat:
```python
In [1]: from datetime import datetime

In [2]: datetime.fromisoformat("2023-07-09T13:47:17.000Z")
Out[2]: datetime.datetime(2023, 7, 9, 13, 47, 17, tzinfo=datetime.timezone.utc)
```

I think we also need to test what we would do with an isoformatted timestamp that has a non-UTC timestamp, for example:
```
>>> datetime.fromisoformat("2024-12-10T15:30:00+02:00")
datetime.datetime(2024, 12, 10, 15, 30, tzinfo=datetime.timezone(datetime.timedelta(seconds=7200)))
```

the correct answer for the +02:00 example above should be the naive datetime object:
```
datetime(2024, 12, 10, 13, 30)
```

lithium323 (changes requested):

lithium323 (approved):

PR #1141, opened 2024-12-10
Add ETA calculation to logger

dioptx:
For longer running multi processing tasks it would be useful to compute an ETA alongside the existing logger.
This PR aims to implement that.

### DIFF
--- a/src/op_analytics/coreutils/logger.py
+++ b/src/op_analytics/coreutils/logger.py
@@ -5,7 +5,6 @@ import orjson
 import structlog
 from structlog.contextvars import bind_contextvars, bound_contextvars, clear_contextvars
 from structlog.typing import EventDict
-from op_analytics.coreutils.time import now
 from op_analytics.coreutils.env.aware import current_environment, is_k8s
 
 CURRENT_ENV = current_environment().name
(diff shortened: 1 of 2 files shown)


src/op_analytics/coreutils/logger.py
```diff
@@ -109,3 +109,37 @@ def human_interval(num_seconds: int) -> str:
...
+    """
+
+    def __init__(self, total_tasks: int):
+        self.total_tasks = total_tasks
+        self.completed_tasks = 0
+        self.start_time = now()
+
+    def wrap(self, func, current_index: int):
+        def wrapper(target_item):
+            self.completed_tasks += 1
```
lithium323, line 129: the estimate should be done after the function has returned, and maybe only reported if you have already completed more than 3 tasks.

src/op_analytics/coreutils/logger.py
```diff
@@ -109,3 +109,37 @@ def human_interval(num_seconds: int) -> str:
...
+            elapsed = (now() - self.start_time).total_seconds()
+            avg_time_per_task = elapsed / self.completed_tasks
+            remaining_tasks = self.total_tasks - self.completed_tasks
+            eta = remaining_tasks * avg_time_per_task
+
+            # Bind progress and ETA as context vars, so when func logs, these fields appear
+            with bound_contextvars(
+                target_id=f"{current_index:03d}/{self.total_tasks:03d}",
```
lithium323, line 137: looking at this I don't know if `target_id` was the best choice for the name of this information. Maybe something like `progress`? 

Also we could do without the extra info (singe log lines get harder to eyeball the more things they have) so adding only `eta` would be my suggestion.

src/op_analytics/coreutils/threads.py
```diff
@@ -1,13 +1,52 @@
...
+            with bound_contextvars(
+                target_id=f"{current_index:03d}/{self.total_tasks:03d}",
+                # completed=self.completed_tasks,
+                # total=self.total_tasks,
+                # elapsed=f"{elapsed:.1f}s",
+                eta=f"{self.eta:.1f}s" if self.eta is not None else "N/A",
+            ):
+                return func(target_item)
+
+        return wrapper
```
lithium323, line 47:
```suggestion
    def __init__(self, total_tasks: int):
        self.total_tasks = total_tasks
        self.completed_tasks = 0
        self.start_time = now()

    def counter(self, current_index: int) -> str:
        return f"{current_index:03d}/{self.total_tasks:03d}"

    def eta(self) -> str | None:
        if self.completed_tasks < 3:
            return None

        elapsed = (now() - self.start_time).total_current_indexseconds()
        seconds_per_task = elapsed / self.completed_tasks
        remaining_tasks = self.total_tasks - self.completed_tasks
        eta_seconds = remaining_tasks * seconds_per_task
        return human_interval(eta_seconds)

    def wrap(self, func, current_index: int):
        def wrapper(target_item):
            # Bind progress and ETA as context vars, so when func logs, these fields appear
            with bound_contextvars(counter=self.counter(), eta=self.eta()):
                return func(target_item)

            self.completed_tasks += 1

        return wrapper
```

PR #1202, opened 2025-01-03
Setup clickhouse ingestion for growthepie data

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**


lithium323:
FYI: @chuxinh  this is the implementation that ingests growthepie data from GCS to Clickhouse.  It runs as part of the github action for growthepie.  For now I only set up the `chains_daily_fundamentals_v1` table in clickhouse. Still pending to ingest the metadata table.

PR #1161, opened 2024-12-12
Build Enriched Traces & Transactions + Aggregation Tables

MSilb7:
Working PR - See overall project #1055

closes #1160 - Transaction Models Exploration
closes #1168 - Trace Models Exploration

The end results of these models are also expected to be used in conjunction with: #1151 

TODO:
- [x] ~Enriched Transactions~ Extend existing transaction_fees model
  - [ ] Validate that daily_address_summary sustains
- [x] Enriched Traces
- [x] Aggregations for Transactions
- [x] Aggregations for Traces
- [ ] Build Tests

Aggregation strategy:
- Make a view that aggregates at the transaction level (at least), so we can preserve transaction-level fields (i.e. gas usage).
- Then build all later aggregations from there.
- For aggregations below the transaction level, make sure we don't sum up transaction-level fields, since they're not part of the whole.

I'm the worst at excalidraw, but this is what I envision. Progress moves left to right, this PR ends at "Aggregation Models"
image

PR #1219, opened 2025-01-06
Updates to daily address summary -> move to v2

lithium323:
These updates are picked up from #1161 where Michael introduced refined traces and some modifications to refined transactions.  

I don't think we are yet using the daily address summary_v1 anywhere, so I think it's ok to move to v2.  

At the end of the day we might not even use v2 in its current form, since it might be better to build the daily aggregate on top of the refined traces and refined txs datasets that are being introduced in #1161 . 

That said, I thought it would be good to save the changes from Michael because we may need to replicate this if we rebuild on top of the refined txs dataset. 

**Description**

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/src/op_analytics/datapipeline/models/code/daily_address_summary.py
+++ b/src/op_analytics/datapipeline/models/code/daily_address_summary.py
@@ -10,7 +10,7 @@ from op_analytics.datapipeline.models.compute.types import NamedRelations
         "ingestion/transactions_v1",
     ],
     expected_outputs=[
-        "summary_v1",
+        "summary_v2",
     ],
     auxiliary_views=[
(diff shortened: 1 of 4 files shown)


MSilb7 (approved):
weeeee

chuxinh:
yay! does the refined transactions table needs to be merged before this can run??

lithium323:
> yay! does the refined transactions table needs to be merged before this can run??

at the moment this model uses refined transactions as an auxiliary view,  so it reuses the SQL and not the materialized table.  Once #1161  is up and running we will have a refined transactions dataset that we can use as an input.

PR #1222, opened 2025-01-07
Initial version of github metrics tables and updates to legacy bigquery tables to avoid windowing issues

dioptx:
In addition to setting up an initial prototype for new github metrics I'm also using this PR to move github to dagster scheduling.  Last week we migrated from github actions to circle ci and realized circle's interface is not very friendly for our data scheduling workflows. 

It's a big PR, unfortunately needed to address many things at once. 

### Functionality updates

- Fixed windowing issues with github traffic data pulls.
- Introduced the PyGithub library to use as the backbone for the github activity data pulls (PRs, Issues, Comments, Reviews).  This client library is battle tested so will be easier to maintain than reimplementing ourselves.

## Data Pull Mechanics

- Each day we take a snapshot of open PRs and Issues.
- We also take a snapshot of PRs and Issues closed on the most recent N days. 
- For the PRs fetched we also fetch comments and reviews.

The resulting tables will have a lot of redundant data because open PRs that don't change will keep getting fetched for as long as they are open.  Also PRs and Issues that are closed will be fetched and included in the snaphsot for N days after they are closed. 

[IMPORTANT] New comments on already merged PRs may occur but they are out of scope for this effort since no metrics rely on this information.

Open Question for Dennis, how were you achieving pulling only once in the initial pull:
- When a PR is approved & merged, all metrics for its lifecycle can be calculated so there's no need to keep checking it every day. We only need to do it once in the initial pull. 

### Pending Work

- Have not yet added pulling for commits or releases. 
- This PR does not do any data processing. Calculates metrics like time-to-approval and time-to-merge for each PR should be done on a separate PR as a daily job downstream of data fetching.


lithium323:
thanks so much for this head start work Dennis! I started looking at the branch and will push some commits to it with my changes. I want to keep the old github code that goes to bigquery so we can have some overlap between old and new as we migrate.

PR #1235, opened 2025-01-09
Defillama token tvl model

chucktraverse:
This PR enriches the defillama token protocol breakdown data with some filters required for removing double counted TVL and identifies misrpresented tokens. This data will be stored in GCS and updated daily. A BQ table will then be made and a view will be created that joins the correct metadata and filters can be subsequently used in downstream processes.


src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,291 @@
...
+    Container for processed DeFiLlama TVL breakdown data.
+
+    Attributes:
+        df_tvl_breakdown: DataFrame containing protocol token TVL data
+    """
+
+    df_tvl_breakdown: pl.DataFrame
+
+
+def process_defillama_data() -> DefillamaTVLBreakdown:
```
lithium323, line 49: since the `process_defillama_data` function exists exclusively to build the TVLBreakdown dataframe you could make it a classmethod so that at the callsite it is obvious what it is meant to do

So at the callsite it would look something like:
```
DefillamaTVLBreakdown.of_date("2025-01-10")
```

And as a class method this would be:
```
    @classmethod
     def of_date(cls, ctx: DuckDBContext, config: DefiLlamaConfig, datestr: str):
        <ALL OF THE LOGIC GOES HERE>

        return cls(df_tvl_breakdown=df_tvl_breakdown)
```

You are going to need a way to parameterize the process for a given date, as opposed to using datetime.now() in the function.   It's also a good idea for the function to receive the duckdb context and the config instead of being responsible for initializing them within.

chucktraverse: Do you mean making it a class method within the DefillamaTVLBreakdown class?

lithium323: yeah that's right, so to build it you call that `of_date()` "factory" method on the class itself.  At the callsite that looks neat because you know the class of the object that you are putting together.

chucktraverse: Cool, implementing it now

chucktraverse: This of_date will work for daily uploads but what about backfills?

src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,295 @@
...
+    alignment_df: pl.DataFrame
+    token_categories: pl.DataFrame
+
+
+@dataclass
+class DefillamaTVLBreakdown:
+    df_tvl_breakdown: pl.DataFrame
+
+    @classmethod
+    def of_date(cls, datestr: str, config=DefiLlamaConfig):
```
lithium323, line 42: looking at it again I think we still need to "deconstruct" the process a little bit more so it is more friendly to scheduling it.  But the exact way to do that could be better done when we start working on the scheduling part. So I would say it's ok to merge the PR for now and we can work on scheduling and related changes separately.

src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,295 @@
...
+        config = load_config(
+            "/Users/chuck/codebase/op-analytics/src/op_analytics/datapipeline/models/code/defillama_config.yaml"
+        )
+
+        # Load data
+        view1 = DefiLlama.PROTOCOLS_TOKEN_TVL.read(min_date=three_days_ago, max_date=datestr)
+        view2 = DefiLlama.PROTOCOLS_METADATA.read(min_date=one_day_ago)
+
+        # Process protocol TVL
+        df_protocol_tvl = pl.from_pandas(
```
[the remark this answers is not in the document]
chucktraverse: whoops! Forgot to undo that!

src/op_analytics/datapipeline/models/code/defillama_config.yaml
```diff
@@ -0,0 +1,217 @@
...
+  - "borrowed"
+  - "staking"
+  - "pool2"
+  - "polygon-bridge-&-staking"
+
+categories_to_filter:
+  - "CEX"
+  - "Chain"
+
+alignment_dict:
```
MSilb7, line 20: We have the chain distinctions already in the chains metadata view in BQ. So. maintaining this twice is unnecessary.

src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,291 @@
...
+        client = ctx.client
+        date = datetime.strptime(datestr, "%Y-%m-%d")
+        one_day_ago = (date - timedelta(days=1)).strftime("%Y-%m-%d")
+        three_days_ago = (date - timedelta(days=3)).strftime("%Y-%m-%d")
+        config = load_config(
+            "/Users/chuck/codebase/op-analytics/src/op_analytics/datapipeline/models/code/defillama_config.yaml"
+        )
+
+        # Load data
+        view1 = DefiLlama.PROTOCOLS_TOKEN_TVL.read(min_date=three_days_ago, max_date=datestr)
```
MSilb7, line 67: should we name these views something, so that it's easier to follow the code?

MSilb7 (commented):
Overall Comment:
1. I'd expect the outcome of this to be an "enriched defillama tvl" that shows me true/false fields like "is_misrepresented", "is_cex", "is_double_count", etc that help us filter.
2. The token mappings should be a separate isolated table, rather than joined here. We should apply token-mappings at the query level, so that we don't have to re-backfill every time we make a new mapping.
3. The chain mappings already exist in metadata tables & views, so we don't need to duplicate that here.

src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,291 @@
...
+        view2 = DefiLlama.PROTOCOLS_METADATA.read(min_date=one_day_ago)
```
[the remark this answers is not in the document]
chucktraverse: In production it'll effectively be one day ago because the function takes a date string as an input and it'll default to the run date.

src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,291 @@
...
+                END AS misrepresented_tokens
+            FROM {view2}
+        """).pl()
+
+        # Merge and process data
+        df_all = df_protocol_tvl.unique().join(df_metadata.unique(), on="protocol_slug", how="left")
+
+        # Join mappings and process data fields
+        df_all = df_all.join(config.alignment_df, on="chain", how="left")
+        df_all = df_all.join(config.token_categories, on="token", how="left")
```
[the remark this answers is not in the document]
chucktraverse: Removed and will add them as gsheet --> BQ tables

src/op_analytics/datapipeline/models/code/defillama_filter_logic.py
```diff
@@ -0,0 +1,291 @@
...
+        [
+            {"chain": chain, "alignment": alignment}
+            for chain, alignment in config["alignment_dict"].items()
+        ]
+    )
+
+    # Create token categories DataFrame from token data
+    token_categories = pl.DataFrame(config["token_data"])
+
+    return DefiLlamaConfig(
```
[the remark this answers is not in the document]
chucktraverse: Filtering removed and changed to a "to_filter_out" column similar to your defillama_utils.py file. Happy to talk about breaking out the filters into more granular columns if that's desired

src/op_analytics/datapipeline/models/code/defillama_config.yaml
chucktraverse: so we have `is_protocol_misrepresented` and `to_filter_out` now. If we want `is_cex` and `is_double_counted` it would be helpful to talk through exactly which filters correspond to those. It's an easy change, just need to know which contribute to double counting.

For `is_cex`, is it simply if the protocol_slug or chain ends in "-cex"?

chucktraverse:
@lithium323 I'm done with the logic. However I'm not sure how we want to schedule this. I added it to app.py in pulls but LMK if there's a different way you want to do it.

 I also still don't understand how this data will be fully backfilled in GCS. I tried doing a backfill but even with `OPLABS_ENV=prod` it wrote locally

PR #1237, opened 2025-01-10
Upload Adhoc notebooks

dioptx:
**Description**

Upload Ad-hoc Notebooks for Github, TVL & User balance data


lithium323 (approved):
Looks great! It's good to save this code so others can leverage it!

One comment on the use of the API token in the notebook. I would advocate for using `os.environ` as a way to get the api token.  If we have a place in a cell where we can paste it then we are at risk of saving and pushing a commit with the credential.

lithium323 (changes requested):
Now that I am working on the github metrics PR I see that on the github notebooks here we have a lot of overlap with the functions introduced to ingest raw metrics. 

I think our goal should be to first merge the functionality into our library and then update the notebooks to use the methods from the library.

PR #1252, opened 2025-01-15
Read from ethereum-lists/chains

Abuchtela:
Add code to read from ethereum-lists/chains for EVM chain metadata.

* Filter out non-EVM chains like Bitcoin or Solana.
* Integrate the fetched data with the existing chain metadata structure.
* Handle multiple L1 cost address combinations and manually inputted metadata.
* Read private data from CRM for a private extension to chain metadata.
* Build the "chain dashboard" using the fetched data.


lithium323:
spam

PR #1264, opened 2025-01-16
Add erc20 / erc721 transfers

chuxinh:
#1134 

**Description**
- Add erc20 and erc721 transfers 
- Add unit tests


MSilb7:
This isn't using the ABI right? So we still need to build that for future models.

chuxinh:
> This isn't using the ABI right? So we still need to build that for future models.

Correct, we're not using ABI for that given it's rather standard. Discussed with @lithium323 and we don't wanna be blocked by ABI for this. He might make changes in the future :D

src/op_analytics/datapipeline/models/templates/token_transfers.sql.j2
```diff
@@ -0,0 +1,20 @@
...
+    ,l.address AS contract_address
+    ,case when array_length(indexed_args) = 2 then l.data.hex_to_lossy() else null end AS amount
+    ,case when array_length(indexed_args) = 2 then l.data.hex_to_lossless() else null end AS amount_lossless
+    ,indexed_event_arg_to_address(indexed_args[1]) AS from_address
+    ,indexed_event_arg_to_address(indexed_args[2]) AS to_address
+    ,hex_to_lossless(indexed_args[3]) AS token_id
+FROM {{ raw_logs }} AS l
+WHERE
+    true
+    AND l.topic0 LIKE '0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef%' -- transfers
```
lithium323, line 20: this can be `=` instead of `LIKE`

lithium323 (approved):

lithium323 (approved):

Issue #1265, opened 2025-01-16
ERC20 Further Enhancement Workstream


chuxinh:
Extended work from #1134 after #1264 being merged.

### Superchain Token Metadata ✅ 
- Pull data and get constant updates from [optimism.tokenlist.json](https://github.com/ethereum-optimism/ethereum-optimism.github.io/blob/master/optimism.tokenlist.json)
- There's existing job that's pulling data from the list: #1266 
- Added api pipeline here #1267 

### Enrich Contract Creation with ERC20 and ERC721 contract mapping 
- Identify if a contract created is ERC20 or ERC721
- Question if we should just create separate token contract creation list as we might want to add additional token specific categorisation (Superchain ERC20, OFT, NTT etc)
- Enabling us to know the % of Superchainerc20 and interop assets v.s. the rest of token assets on Superchain

### Token Standard Identification
- Pipelining the logic behind [SuperchainERC20 Deployments](https://app.hex.tech/61bffa12-d60b-484c-80b9-14265e268538/app/66ac2029-23ab-4b0d-b9a8-51308dad0941/latest?) for Superchain ERC20 tokens
- Identify NTT: https://wormholescan.io/#/analytics/ntt
- Identify OFT
- Other tokens to consider: Axelar’s Interchain Token Service, Hyperlane’s Warp Token, xERC20
- Use the above to help us segment TVL by token standards

### Token Price Data
- TODOS: check Defillama API (https://defillama.com/docs/api)

PR #1293, opened 2025-01-24
Add erc20 Contract Creation

chuxinh:
Related to #1265 
---
- Use bytecode to identify if a contract is erc20 
- Create separate logic to identify if the token type is Superchain erc20 

I'm expecting other logic such as OFT and NTT to follow, how flexible it is if there are future addition of sql template added to this pipeline?


src/op_analytics/datapipeline/models/templates/erc20_contracts/superchainerc20.sql.j2
```diff
@@ -0,0 +1,36 @@
+WITH create2 AS (
+  SELECT
+    *
+    , 'SuperchainERC20' AS token_type
+  FROM {{ contract_creation_traces }}
+  WHERE
+    -- crosschainBurn(address,uint256): 0x2b8c49e3
+    -- crosschainMint(address,uint256): 0x18bf5077
+    (output LIKE '%2b8c49e3%' AND output LIKE '%18bf5077%')
```
MSilb7, line 9: Should we rename all of this to ERC____ (whatever the crosschainMint/Burn standard is?)

Since we actually need to call these functions on the contract to check the address with mint/burn rights to confirm if it's SuperchainERC20

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
```diff
@@ -0,0 +1,14 @@
...
+  , 'erc20' AS token_standard
+  , sc.token_type
+FROM {{ contract_creation_traces }} AS ct
+LEFT JOIN {{ superchainerc20 }} AS sc
+  ON
+    ct.chain_id = sc.chain_id
+    AND ct.contract_address = sc.contract_address
+    AND ct.network = sc.network
+WHERE
+  -- transfer(address,uint256): 0xa9059cbb
```
MSilb7, line 12: Nice approach, some things to check on:

1. Does this catch tokens created behind a proxy (like USDC https://optimistic.etherscan.io/token/0x0b2c639c533813f4aa9d7837caf62653d097ff85#readContract)? This is useful for the underlying contracts at least, but we may need to use transfers to get the tokens themselves.

^ It seems like you get this in your approach by calling this 'erc20_contract_creation'. So this isn't the tokens table (yet) but it could be a contributor.

2. Are there other functions in the [ERC20 standard](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#IERC20) that we want to check for? Like approve()

We can ask protocol/eng if this approach is sensible as well, or if there's nuance to look out for (like do we need PUSH0 prepending, or something else)

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
```diff
@@ -0,0 +1,14 @@
+SELECT
+  ct.*
+  , 'erc20' AS token_standard
+  , sc.token_type
```
MSilb7, line 4: Should we coalesce this to some other status, or keep it as null if not superchain erc20?

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
MSilb7: We'll also need to load in a static/manual list of pre-regenesis tokens that won't have a creation trace on OPM.

We have that somewhere.

MSilb7 (commented):
Left some comments, just to make sure we're pulling what we think we are.

Also note that before backfilling this, we should statically add the OVM1 tokens that won't have creation events.

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
```diff
@@ -0,0 +1,14 @@
...
+FROM {{ contract_creation_traces }} AS ct
+LEFT JOIN {{ superchainerc20 }} AS sc
```
MSilb7, line 6: I wonder if it makes sense to join to token standards here, or as a later view?

Since we can add more of these over time.

src/op_analytics/datapipeline/models/templates/erc20_contracts/superchainerc20.sql.j2
```diff
@@ -0,0 +1,36 @@
...
+      AND c2.chain_id = c0.chain_id
+      AND c2.block_number = c0.block_number
+      AND c2.transaction_hash = c0.transaction_hash
+      AND c2.contract_address = c0.tr_from_address
+      AND c0.trace_type LIKE 'create%'
+  WHERE
+    c2.trace_type = 'create2'
+    AND (c0.output LIKE '%2b8c49e3%' AND c0.output LIKE '%18bf5077%')
+    -- create3: https://gist.github.com/Aboudjem/3ae8cc89ebff8c086e9b9111b1d06e6d#file-create3-sol-L25
+    AND (c2.output LIKE '%0x363d3d37363d34f0%')
```
lithium323, line 29: I'm curious about the 0x in `LIKE '%0x363d3d37363d34f0%'`.  

Doesn't 0x always appear at the start only? If so, shouldn't it be `LIKE '0x363d3d37363d34f0%'`

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
chuxinh: i'm thinking of adding other `token_type` like NTT, OFT later after getting to understand the logic to fetch those better. But I assume most are just normal erc20 so maybe we should leave it as null?

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
chuxinh: yea indeed, currently keeping it flexible as a separate view, but i wonder how many that'd be. Or maybe in the future we just keep creating different token types in template and then left join all in dbt instead of the current pipeline. Also cc: @lithium323 if you have any thought here

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc20_contract_creation.sql.j2
chuxinh: I think the issue with the current pipeline is that it's by blockbatch, so technically the logic won't be able to look back at the entire history (this is where dbt could come in) 
- given the constraint of using the bytecode to identify erc20 in the above example, and if we want to use joining `creation_traces` + `erc20_transfers`
- that means we need to create erc20 token first transfers with contract_address and it's min(block_timestamp) and then join to contract creation to see if there's a match

cc: @lithium323 if the above understanding is correct

src/op_analytics/datapipeline/models/templates/erc20_contracts/superchainerc20.sql.j2
chuxinh: yea can do that. 
note another contract call necessary here to call and verify the actual mint and burn address
it's erc7802: https://docs.optimism.io/stack/interop/superchain-erc20

also lemme follow up with the product team next week if it's even possible to know the contract address at all

src/op_analytics/datapipeline/models/templates/erc20_contracts/erc7802.sql.j2
```diff
@@ -0,0 +1,36 @@
...
+  WHERE
+    -- crosschainBurn(address,uint256): 0x2b8c49e3
+    -- crosschainMint(address,uint256): 0x18bf5077
+    (output LIKE '%2b8c49e3%' AND output LIKE '%18bf5077%')
+)
+
+, create3 AS (
+  SELECT
+    c2.*
+    , 'erc7802' AS token_type
```
lithium323, line 15: do we need an additional column so that we can identify which part of the UNION ALL each row came from?

PR #1299, opened 2025-01-28
Implement Github Metrics

dioptx:
This PR adds a new compute_pr_metrics function that merges GitHub PR, comments, and reviews data into a daily time series of Development related metrics.


src/op_analytics/datasources/github/metrics/compute.py
```diff
@@ -0,0 +1,222 @@
...
+
+def compute_pr_metrics(
+    prs_df: pl.DataFrame,
+    comments_df: pl.DataFrame,
+    reviews_df: pl.DataFrame,
+) -> pl.DataFrame:
+    """
+    Compute daily GitHub PR metrics using three dataframes (PRs, Comments, Reviews).
+
+    1) PRs (prs_df)
```
lithium323, line 17: Since the schemas are defined as part of the code, should we reference that definition instead? It will be harder to keep the docs in sync if we copy paste it here in the docstring.

src/op_analytics/datasources/github/metrics/execute.py
```diff
@@ -0,0 +1,49 @@
...
+    1. Load raw data from GCS
+    2. Compute metrics
+    3. Write metrics back to GCS
+    """
+    # 1. Load raw data from GCS with date filters
+    prs_view = Github.PRS.read(min_date=min_date, max_date=max_date)
+    comments_view = Github.PR_COMMENTS.read(min_date=min_date, max_date=max_date)
+    reviews_view = Github.PR_REVIEWS.read(min_date=min_date, max_date=max_date)
+
+    # Convert views to DataFrames using DuckDB
```
lithium323, line 21: This looks like a `read_polars()` utility that we could create on the `DailyDataset` class.

src/op_analytics/datasources/github/metrics/compute.py
```diff
@@ -0,0 +1,222 @@
...
+                (pl.col("is_approved").sum() / pl.count()).alias("approval_ratio"),
+                (pl.col("comment_count").sum() / pl.count()).alias("avg_comments_per_pr"),
+                (pl.col("is_merged").sum() / pl.count()).alias("merged_ratio"),
+                pl.col("user").struct.field("login").n_unique().alias("active_contributors"),
+            ]
+        )
+        .sort(["repo", "dt"])
+    )
+
+    return daily_metrics
```
lithium323, line 222: Could you define the expected schema for daily_metrics as a constant?  And then before returning we can do something like:
```
raise_for_schema_mismatch(
        actual_schema= daily_metrics.schema,
        expected_schema=EXPECTED_DAILY_METRICS_SCHEMA,
    )
```

src/op_analytics/coreutils/partitioned/dailydata.py
```diff
@@ -162,6 +162,33 @@ def create_clickhouse_view(self) -> None:
...
+        directly to a Polars DataFrame.
+
+        Args:
+            min_date: Optional start date filter
+            max_date: Optional end date filter
+            date_range_spec: Optional date range specification
+            location: Data location (GCS or LOCAL)
+
+        Returns:
+            pl.DataFrame: The loaded data as a Polars DataFrame
```
lithium323, line 184: The read_polars() method was already added to main. Found myself needing to use it and remembered you were going to use it as well. It was added last week.

src/op_analytics/datasources/github/dataaccess.py
```diff
@@ -10,12 +10,20 @@ class Github(DailyDataset):
     other third-party datasets (like DefiLlama, etc.).
     """
 
-    # Initial github data pull following Bella's Hex notebook.
-    TRAFFIC_METRICS = "repo_metrics_v1"
-    REFERRER_METRICS = "repo_referrers_v1"
```
lithium323, line 15: I think I might be missing something here. It seems like we shouldn't be deleting the TRAFFIC and REFERRER metrics that we pull from the github api.

src/op_analytics/datasources/github/dataaccess.py
```diff
@@ -10,12 +10,20 @@ class Github(DailyDataset):
...
-
-    # Github data brought in for platform metrics dashboards by Dennis.
+    # Raw github data
     ISSUES = "github_issues_v1"
     PRS = "github_prs_v1"
     PR_COMMENTS = "github_pr_comments_v1"
     PR_REVIEWS = "github_pr_reviews_v2"
+    REPO_METRICS = "github_repo_metrics_v1"
+
+    def write(self, dataframe: pl.DataFrame, sort_by: list[str] | None = None):
```
lithium323, line 20: In a recent `main` commit we made it so all subclasses of DailyData inherit this write method. So we don't need to have it explicitly anymore.

src/op_analytics/dagster/assets/github.py
```diff
@@ -46,11 +45,40 @@ def activity_views():
...
+@asset(deps=[activity], group_name="github", name="repo_metrics")
+def repo_metrics(context: OpExecutionContext) -> None:
+    """Dagster asset to compute and write GitHub PR metrics data.
+
+    This asset depends on the activity asset which pulls raw PR data.
+    Computes and writes daily metrics about PR activity and performance across repos.
+    """
+    from op_analytics.datasources.github.metrics.execute import execute_pull_repo_metrics
+
+    result = execute_pull_repo_metrics()
```
lithium323, line 64: could you run locally `uv run mypy`?  I think there is a problem here because the args in the `execute_pull_repo_metrics` function are not declared as optional at the moment.  

My bad I should have set up a CI job by now so we can get this feedback automatically after opening the PR.

src/op_analytics/datasources/github/metrics/execute.py
```diff
@@ -0,0 +1,64 @@
...
+    log.info(
+        "Loaded raw GitHub data",
+        min_date=min_date,
+        max_date=max_date,
+        prs_count=len(prs_df),
+        comments_count=len(comments_df),
+        reviews_count=len(reviews_df),
+    )
+
+    process_dt = date_tostr(now_date())
```
lithium323, line 38: minor: here you could use `now_dt()`.

src/op_analytics/datasources/github/metrics/execute.py
```diff
@@ -0,0 +1,64 @@
...
+      1. Load raw data from GCS.
+      2. Compute rolling/detailed metrics.
+      3. Write the computed metrics to the REPO_METRICS dataset.
+
+    Args:
+        min_date: Start date for data processing (ISO format)
+        max_date: End date for data processing (ISO format)
+
+    Returns:
+        dict: Summary of processing results.
```
lithium323, line 22: I find that the Args: and Returns: parts of the docstrings are not really that helpful.  We can try to use the docstring to explain more nuanced concepts. For example, what is the relation between the `min_date/max_date` arguments and the `process_dt` that we define below? 

I'm afraid there is a disconnect here. For example, what would it mean for me to pass some really old values of `min_date/max_date` but still write the output under the present date?

src/op_analytics/datasources/github/metrics/compute.py
```diff
@@ -0,0 +1,364 @@
...
+    "unique_contributors": pl.UInt32,
+    "approved_prs": pl.UInt32,
+    "rejected_prs": pl.UInt32,
+    "stale_prs": pl.UInt32,
+    "review_requested_prs": pl.UInt32,
+    "period_type": pl.Utf8,
+}
+
+
+def prepare_dataframes(
```
lithium323, line 33: The prepare_dataframes function has too much scope, so it makes the logic somewhat harder to follow.  

We should isolate concepts a bit more.  So each entity type (prs, commment, and reviews) could be prepared separately.  That way when reading the code one only has one thing in mind instead of 3.

src/op_analytics/datasources/github/metrics/compute.py
```diff
@@ -0,0 +1,364 @@
...
+
+
+def compute_all_metrics(
+    prs_df: pl.DataFrame,
+    comments_df: pl.DataFrame,
+    reviews_df: pl.DataFrame,
+) -> pl.DataFrame:
+    """Compute metrics for multiple rolling window sizes and combine results"""
+
+    prs_df, comments_df, reviews_df = prepare_dataframes(prs_df, comments_df, reviews_df)
```
lithium323, line 315: when we pull data from github we fetch information redundantly, because for each day we load all open PRs plus PRs that have been closed in the last 30 days.  

I'm somewhat concerned in that I don't see a data preparation step that explicitly accounts for that redundancy. I think we should do that to ensure correctness and so our future selves can be more aware of that nuance when we are looking at the compute logic later in the future.

src/op_analytics/datasources/github/metrics/compute.py
```diff
@@ -0,0 +1,214 @@
...
+) -> pl.DataFrame:
+    """Compute metrics for a rolling window of given size"""
+
+    # Filter out bot comments and sort data
+    comments_df = comments_df.filter(
+        pl.col("user").struct.field("login").str.contains(r"(?i)\[bot\]$").not_()
+    )
+    prs_df = prs_df.sort("created_at")
+    comments_df = comments_df.sort("created_at")
+    reviews_df = reviews_df.sort("submitted_at")
```
lithium323, line 98: can filtering and sorting be moved outside?  Looks like we are repeating the operation every time we call `_compute_rolling_for_interval`

src/op_analytics/datasources/github/metrics/cohort.py
```diff
@@ -0,0 +1,52 @@
...
+
+def compute_cohort_metrics(prs_df: pl.DataFrame, period_str: str) -> pl.DataFrame:
+    """
+    Compute cohort metrics based on PR creation date.
+    """
+    # Pure python implementation of the requested_reviewers_count column.
+    reviewers_col = prs_df["requested_reviewers"].to_list()
+    counts = [len(item) if isinstance(item, list) else 0 for item in reviewers_col]
+    if "requested_reviewers_count" in prs_df.columns:
+        prs_df = prs_df.drop("requested_reviewers_count")
```
lithium323, line 15: I see that there is a different approach for two similar things.  Here we add the `requested_reviewers_count` feature for a PR which is used for computing a metric.  This is similar to `earliest_non_bot_comment_at` which is something that we add to each PR as a feature.  

The difference is that for `earliest_non_bot_comment_at` we apply the logic once before computing metrics, whereas for `requested_reviewers_count` we apply the logic inside the `compute_cohort_metrics` function which means the computation is repeated for every period aggregation. 

I think it would be good to consolidate all pre-processing logic in a separate module.

src/op_analytics/datasources/github/metrics/cohort.py
```diff
@@ -0,0 +1,52 @@
...
+                .median()
+                .alias("median_time_to_first_non_bot_comment_hours"),
+                ((pl.col("merged_at") - pl.col("created_at")).dt.total_hours())
+                .median()
+                .alias("median_time_to_merge_hours"),
+                pl.col("user").struct.field("login").n_unique().alias("unique_contributors"),
+                # Turn any row with requested_reviewers_count > 0 into 1, then sum them.
+                pl.when(pl.col("requested_reviewers_count") > 0)
+                .then(pl.lit(1))
+                .otherwise(pl.lit(0))
```
lithium323, line 44: maybe you could have added the feature called `has_requested_reviewers` instead of  `requested_reviewers_count` so the aggregation doesn't also have to include feature logic.

lithium323, line 44: here we could add a feature called `has_requested_reviewers` instead of  `requested_reviewers_count` so the aggregation doesn't also have to include feature logic.

src/op_analytics/datasources/github/metrics/cohort.py
```diff
@@ -0,0 +1,52 @@
...
+        .agg(
+            [
+                pl.col("pr_number").n_unique().alias("new_prs"),
+                ((pl.col("approved_at") - pl.col("created_at")).dt.total_hours())
+                .median()
+                .alias("median_time_to_first_review_hours"),
+                ((pl.col("earliest_non_bot_comment_at") - pl.col("created_at")).dt.total_hours())
+                .median()
+                .alias("median_time_to_first_non_bot_comment_hours"),
+                ((pl.col("merged_at") - pl.col("created_at")).dt.total_hours())
```
lithium323, line 37: Left a similar comment below, but I think you can consider the math here a feature that you can add on each PR separately before you enter the aggregation step.  That way it is going to be more clear how it is computed.  

So you would have some features that you compute only once before entering the aggregation over different time periods:
```
time_to_first_review_hours
time_to_first_non_bot_comment_hours
```

It would also be good to have those outside because they can be closer to the logic that defines some of the qualifiers. So the definition of `time_to_first_non_bot_comment_hours` would be right beside the code that defines what constitutes a `non_bot_comment`.

PR #1302, opened 2025-01-28
Update defillama pulls to have larger overlap (and save backfill notebook)

lithium323:
When new protocols get added by Defillama there can be backdated new data added. By extending the overlap window we expect to pick up more of that data. 

**Description**

**Tests**

**Additional context**

**Metadata**


src/op_analytics/datasources/defillama/volume_fees_revenue/helpers.py
```diff
@@ -35,12 +35,9 @@ def chain_endpoint(chain: str, data_type: ENDPOINT_TYPES):
     return f"https://api.llama.fi/overview/{endpoint}/{quoted_chain}?excludeTotalDataChart=false&excludeTotalDataChartBreakdown=false&dataType={data_type}"
 
 
-TABLE_LAST_N_DAYS = 3  # upsert only the last X days of volume fetched from the api
```
lithium323, line 38: this wasn't being used

PR #1305, opened 2025-01-28
Create external tables and views for defillama datapipeline

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**

### DIFF
--- a/src/op_analytics/datapipeline/etl/bigqueryviews/ddl/dailydata_defillama/defillama_tvl_breakdown_filtered.sql
+++ b/src/op_analytics/datapipeline/etl/bigqueryviews/ddl/dailydata_defillama/defillama_tvl_breakdown_filtered.sql
@@ -1,0 +1,22 @@
+SELECT
+  tvl.dt
+  , tvl.chain
+  , COALESCE(md.alignment, 'Other') AS alignment
+  , tvl.protocol_category
+  , tvl.parent_protocol
+  , tvl.protocol_name
+  , tvl.protocol_slug
+  , tvl.is_double_counted
+  , tvl.is_protocol_misrepresented
+  , COALESCE(tm.token_category, 'Other') AS token_category
+  , tvl.token
+  , tm.project AS token_source_project
+  , tm.source_protocol AS token_source_protocol
+  , tvl.app_token_tvl
+  , tvl.app_token_tvl_usd
+FROM `oplabs-tools-data.dailydata_defillama.protocol_token_tvl_breakdown_v1` tvl 
+LEFT JOIN `oplabs-tools-data.defillama.token_mappings` tm 
+  ON tvl.token = tm.token
+LEFT JOIN `oplabs-tools-data.api_table_uploads.op_stack_chain_metadata` md 
+  ON tvl.chain = md.display_name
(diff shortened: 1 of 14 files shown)


src/op_analytics/datapipeline/etl/bigqueryviews/ddl/dailydata_defillama/defillama_tvl_breakdown_filtered.sql
```diff
@@ -0,0 +1,22 @@
...
+  , COALESCE(tm.token_category, 'Other') AS token_category
+  , tvl.token
+  , tm.project AS token_source_project
+  , tm.source_protocol AS token_source_protocol
+  , tvl.app_token_tvl
+  , tvl.app_token_tvl_usd
+FROM `oplabs-tools-data.dailydata_defillama.protocol_token_tvl_breakdown_v1` tvl 
+LEFT JOIN `oplabs-tools-data.defillama.token_mappings` tm 
+  ON tvl.token = tm.token
+LEFT JOIN `oplabs-tools-data.api_table_uploads.op_stack_chain_metadata` md
```
MSilb7, line 20: Note that we'd ~likely modify this to some other source (i.e. gsheet + registry) over time

src/op_analytics/datapipeline/etl/bigqueryviews/ddl/dailydata_defillama/defillama_tvl_breakdown_filtered.sql
```diff
@@ -0,0 +1,22 @@
...
+  , COALESCE(md.alignment, 'Other') AS alignment
+  , tvl.protocol_category
+  , tvl.parent_protocol
+  , tvl.protocol_name
+  , tvl.protocol_slug
+  , tvl.is_double_counted
+  , tvl.is_protocol_misrepresented
+  , COALESCE(tm.token_category, 'Other') AS token_category
+  , tvl.token
+  , tm.project AS token_source_project
```
MSilb7, line 13: This being a "view" means that we don't have to backfill when we add/modify token mappings right?

MSilb7 (commented):
This new view seems to align with what I expect from the current `oplabs-tools-data.temp.defillama_tvl_breakdown_filtered` table, so this works

src/op_analytics/datapipeline/etl/bigqueryviews/ddl/dailydata_defillama/defillama_tvl_breakdown_filtered.sql
lithium323: yeah that's right

src/op_analytics/datapipeline/etl/bigqueryviews/ddl/dailydata_defillama/defillama_tvl_breakdown_filtered.sql
lithium323: I'll add a comment for that on the view

lithium323:
Will go ahead and merge so the steps here run on tonight's dagster run. If we need changes to the view we can look at it as a separate PR.

PR #1322, opened 2025-02-03
Transformations for interop dashboard

lithium323:
**Description**

SQL Transformations to monitor interop contract creation

**Tests**

**Additional context**

**Metadata**


src/op_analytics/transforms/interop/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Find how many tokens of different bridgeable types are created each day. 
+
+
+Methodology
+-----------
+
+We consider 3 token types:
+
+- SuperchainERC20
```
chuxinh, line 15: Maybe we should leave it as ERC7802 here given superchainerc20 still needs to find out give the crosschain mint and burn are given to a specific bridge address

src/op_analytics/transforms/interop/README.md
```diff
@@ -0,0 +1,72 @@
...
+belong to each of the three different token types.
+
+### Contract Creation Traces
+
+We use traces to find out when a token contract was created. 
+
+For NTT and OFT we inner join with the token contract address dimension table to keep only the
+create traces for those token contracts.
+
+For SuperchainERC20 contracts it is more nuanced. The only events we have are transfer events, which
```
chuxinh, line 37: Same for here as well. I think once we are sure of it, in the future we can even create more specific subset of Superchainerc20

src/op_analytics/transforms/interop/update/06_fact_erc7802_create_traces_v1.sql
```diff
@@ -0,0 +1,123 @@
...
+  , tx_from_address
+  , 'create0/3' AS deployment_type
+  , contract_address
+  , trace_address
+  , trace_type
+  , gas
+  , gas_used
+FROM create3
+
+UNION ALL
```
chuxinh, line 101: i remember in the original query, there was a left join with where condition to make sure there's no overlap in the contract address in either `create0/3` or `create2`, wonder if that's not needed here?

```sql
  SELECT c3.* FROM create3 c3
  UNION ALL
  SELECT c2.* FROM create2 c2
      -- make sure no dupes
      LEFT JOIN create3 c3j
          ON c3j.chain = c2.chain
          AND c3j.contract_address = c2.contract_address
      WHERE c3j.chain IS NULL -- no match
```

src/op_analytics/transforms/interop/parking/07_export_token_create_traces_v1.sql
```diff
@@ -0,0 +1,103 @@
...
+  , transaction_index
+  , tr_from_address
+  , tx_from_address
+  , contract_address
+  , trace_address
+  , trace_type
+  , gas
+  , gas_used
+FROM
+  transforms_interop.fact_erc7802_create_traces_v1
```
chuxinh, line 32:
```suggestion
 `blockbatch.contract_creation__create_traces_v1`
```

src/op_analytics/transforms/interop/parking/07_export_token_create_traces_v1.sql
```diff
@@ -0,0 +1,103 @@
...
+WHERE
+  dt >= '2024-10-01'
+  AND (chain_id, contract_address)
+  IN (SELECT
+    chain_id
+    , contract_address
+  FROM transforms_interop.dim_ntt_contract_address_v1)
```
chuxinh, line 39:
```suggestion
  FROM transforms_interop.fact_erc7802_create_traces_v1)
```

src/op_analytics/transforms/interop/parking/07_export_token_create_traces_v1.sql
```diff
@@ -0,0 +1,103 @@
+/**
```
chuxinh, line 1: I'm wondering instead of creating a small subset of just interop tokens created, maybe we can get the 

erc20 created with joins to erc20 first transfers and then left join to the three dim tables with a new column name `token_type`? 

Tho I think the only problem would be that it might omit some of erc20 tokens created before 2024-10-1 (if it's never been transferred again) given we are only starting to compute the first transfers from that date (but I think that should be fine for now and going forward given the purpose of this exercise here).

And any token created before 2024-10-1 that are still being transferred will be recorded still by joining it back to `create_trace`.

src/op_analytics/transforms/interop/parking/07_export_token_create_traces_v1.sql
lithium323: oops I did not mean to commit the 07_export file, that was where I was playing around with the final data that will be exported for the dashboard.  

For 7802, don't we need to keep only the traces that have the crosschainBurn/Mint method ids? (that's what the `fact_erc7802_create_traces_v1` table is doing)

src/op_analytics/transforms/interop/update/06_fact_erc7802_create_traces_v1.sql
lithium323: yeah I checked that it is not needed anymore, that's because I added `AND c0.trace_type != 'create2'` so that the create3 route already excludes any traces that fall under the create2 branch.

PR #1332, opened 2025-02-07
Update to fix some mypy complaints

lithium323:
**Description**

**Tests**

**Additional context**

**Metadata**


src/op_analytics/datasources/github/metrics/compute.py
```diff
@@ -20,33 +20,33 @@
 log = structlog.get_logger()
 
 METRICS_SCHEMA = {
-    "repo": pl.Utf8,
+    "repo": pl.String(),
```
lithium323, line 23: Here the problem is we cannot mix classes (e.g. `pl.Utf8`) with instances (e.g. `pl.Datetime("us")`) in the values of the dictionary. We have either use all classes or use all instances.

PR #1333, opened 2025-02-07
Add new columns to export interop fact tables

chuxinh:
**Description**

- Add some additional columns from `create_traces` to `erc20_create_trace` export


src/op_analytics/transforms/interop/update/06_fact_erc20_create_traces_v1.sql
```diff
@@ -31,6 +48,5 @@ FROM
 WHERE
   dt = { dtparam: Date }
 
-  -- Contract bytecode has the ERC-20 Transfer and Approval events.
-  AND position(output, 'ddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef')
-  AND position(output, '8c5be1e5ebec7d5bd14f71427d1e84f3dd0314c0f7b2291e5b200ac8c7c3b925')
+  -- Contract bytecode has the ERC-20 Transfer events.
```
lithium323, line 51: the comment here should be: 
```
-- Contract address has emitted Transfer events
```

src/op_analytics/transforms/interop/create/fact_erc20_create_traces_v1.sql
```diff
@@ -11,10 +11,21 @@ CREATE TABLE IF NOT EXISTS _placeholder_
...
     `tx_from_address` FixedString(42),
     `contract_address` FixedString(42),
+    `tx_to_address` FixedString(42),
     `trace_address` String,
     `trace_type` String,
     `gas` UInt128,
     `gas_used` UInt128,
+    `value` UInt256,
+    `code` String,
+    `output` String,
```
lithium323, line 21: do we need to keep the `output` column here?

PR #1341, opened 2025-02-10
Add ERC20 daily transfers aggregate to clickhouse

chuxinh:
**Description**
- Adding a summary table for erc20 transfers on daily basis, this can later be used for us to filter tokens for token metadata pull via RPC. 
- Quick check that 75 percentile amounts to around 5k tokens to get metadata, and 90 percentile is around 1.6K. I imagine the initial pull might be large since we have nothing and the RPC call should reduce after the first data pull cc: @lithium323 
- Also adding the `raw_amount` transferred here but I'd like to see a future enriched table with amount in decimals and with volume in USD

I guess this one serves a different purpose v.s. top contracts with what token transferred but lemme know anything is missing @MSilb7


lithium323 (approved):

lithium323 (approved):

Missjones2829 (commented):
Merge

PR #1343, opened 2025-02-11
Expand schema discovery functionality to the dailydata class

dioptx:
Adds two functions that help output the schemas of all the tables that a dailydata class currently has for a specified snapshot date

### DIFF
--- a/src/op_analytics/coreutils/partitioned/dailydata.py
+++ b/src/op_analytics/coreutils/partitioned/dailydata.py
@@ -102,6 +102,43 @@ class DailyDataset(str, Enum):
         print(duckdb_context.client.sql("SHOW TABLES"))
         return view_name
 
+    @classmethod
+    def get_all_datasets(cls) -> dict:
+        """Get all class attributes and their values as a dictionary.
+
+        Returns:
+            dict: Dictionary containing all class attributes and their values
+        """
+        return {
+            name: value
+            for name, value in vars(cls).items()
+            if not name.startswith("_") and isinstance(value, str)
+        }
+
+    @classmethod
+    def infer_all_schemas(cls, datestr: str) -> dict[str, dict]:
+        """Infer Clickhouse schemas for all datasets and return with metadata.
+
+        Args:
+            datestr: Date string to use for schema inference
+
+        Returns:
+            tuple: Tuple containing:
+                - datasets: Mapping of class attribute names to values
+                - schemas: Mapping of dataset names to their inferred schemas
+        """
+        schemas = {}
+        for dataset in cls.all_tables():
+            log.info(f"Inferring schema for {dataset}")
+            try:
+                schema = dataset.infer_clickhouse_schema(datestr)
+                schemas[dataset.value] = schema
+            except IndexError:
+                log.warning(f"No data found for {dataset} on {datestr}, skipping schema inference")
+                continue
+
+        return cls.get_all_datasets(), schemas
+
     def read_polars(
         self,
         min_date: str | date | None = None,


src/op_analytics/coreutils/partitioned/dailydata.py
```diff
@@ -102,6 +102,34 @@ def read(
...
+        Args:
+            datestr: Date string to use for schema inference
+
+        Returns:
+            dict[str, dict]: Mapping of dataset names to their inferred schemas
+        """
+        schemas = {}
+        for table in cls.all_tables():
+            name = table.name
+            value = table.value.split(".")[-1]
```
lithium323, line 118: not sure there are cases where the table.value has a `.` in it, but maybe I'm missing something?

I would suggest using `table` as the key in the `schemas` dict.

src/op_analytics/coreutils/partitioned/dailydata.py
```diff
@@ -102,6 +102,34 @@ def read(
...
+            log.info(f"Inferring schema for {name}")
+            try:
+                schema = table.infer_clickhouse_schema(datestr)
+                schemas[value] = schema
+            except IndexError:
+                log.warning(f"No data found for {name} on {datestr}, skipping schema inference")
+                continue
+            except Exception as e:
+                log.error(f"Error inferring schema for {name}", error=str(e))
+                continue
```
lithium323, line 129: maybe on the docstring, could you add more info about where we would use the `infer_all_schemas` method? I want to understand where it would be used so I can think about the implications of ignoring missing dates or ignoring errors. 

Ate the very least I think we should never eat out an Exception, as we do in line 127.  Maybe the `IndexError` is ok to ignore.

src/op_analytics/coreutils/clickhouse/inferschema.py
```diff
@@ -28,14 +28,12 @@ def parquet_to_subquery(gcs_parquet_path: str, virtual_columns: str = "") -> str
     """
 
 
-def infer_schema_from_parquet(gcs_parquet_path: str, dummy_name: str):
-    """Use a single parquet path to generate a Clickhouse CREATE TABLE dsl statement.
+def get_schema_from_parquet(gcs_parquet_path: str) -> list[dict]:
```
lithium323, line 31: looks good thanks for splitting this out!

src/op_analytics/coreutils/partitioned/dailydata.py
dioptx: I added this split because I want to get the raw string values of the table names to use them later, and the .all_tables() function returns the full description that I assumed is intended

image

src/op_analytics/coreutils/partitioned/dailydata.py
```diff
@@ -105,6 +105,8 @@ def read(
     @classmethod
     def infer_all_schemas(cls, datestr: str) -> dict[str, dict]:
         """Infer Clickhouse schemas for all datasets and return with metadata.
+        This function can be used for ad-hoc analytics & schema inference,
+        user facing data discoverability tools and data validation and testing.
```
lithium323, line 109: thanks for adding the comment.  One thing I was curious is why is the clickhouse schema specifically needed. Clickhouse types have some quirks and are not as commonplace as parquet types. 

If what we need are the types from the parquet file and not necessarily the clickhouse equivalent types then we shouldn't go through the ordeal of creating a temp table in clickhouse to get that.

src/op_analytics/coreutils/partitioned/dailydata.py
lithium323: hey, the all_table() function returns a list with all of the members of the enum. The string representation of the enum member might be where you see a dot. But there is no dot in `value`. 

Screenshot 2025-02-14 at 2 47 24 PM

lithium323: the table enum members are hashable, which means you can use them as keys in the dictionary you are returning.

src/op_analytics/coreutils/partitioned/dailydata.py
```diff
@@ -129,14 +155,48 @@ def read_polars(
...
+    def create_clickhouse_view(self) -> None:
+        # Database used in ClickHouse to store views that point to GCS data.
+        external_db_name = f"dailydata_{self.db}"
+
+        return create_gcs_view(
+            db_name=external_db_name,
+            table_name=self.table,
+            partition_selection="CAST(dt as Date) AS dt, ",
+            gcs_glob_path=f"{self.root_path}/dt=*/out.parquet",
+        )
```
lithium323, line 198: it looks like the PR is bringing back some code that has been deleted over on `main`

src/op_analytics/coreutils/partitioned/dailydata.py
```diff
@@ -129,14 +155,48 @@ def read_polars(
...
 
-    def infer_clickhouse_schema(self, datestr: str):
-        """Use a single parquet file in GCS to infer the Clickhouse schema for the table."""
+    def get_clickhouse_schema(self, datestr: str) -> list[dict]:
+        """Return the parquet schema inferred from a sample parquet file."""
         paths = query_parquet_paths(
             self.root_path, DataLocation.GCS, DateFilter.from_dts([datestr])
         )
 
         infer_schema_from_parquet(paths[0], dummy_name=f"{self.db}.{self.table}")
```
lithium323, line 164: In the `dailydata.py` file I see the `import infer_schema_from_parquet` has been deleted. Here in the code I see it `infer_schema_from_parquet` being used but this usage is undefined since the import was removed.

PR #1361, opened 2025-02-12
Add Token Metadata Prototype

chuxinh:
Just prototyping pulling token metadata from public RPC


src/op_analytics/datasources/chainsrpc/ddl/onchain.dim_erc20_token_metadata_v1.sql
```diff
@@ -0,0 +1,20 @@
+CREATE TABLE IF NOT EXISTS onchain.dim_erc20_token_metadata_v1
+(
+    `chain` String,
```
MSilb7, line 3: given that we have `chain_id`, I wonder if also having the `chain` string is overkill?

Just thinking of handling for if/when a chain name or config might change, chain_id should be the most unchangable? Unless we make sure we don't change one without the other.

src/op_analytics/datasources/chainsrpc/erc20metadata.py
```diff
@@ -0,0 +1,188 @@
...
+            derived.chain_id,
+            derived.contract_address,
+            derived.num_transfers
+        FROM (
+            SELECT
+                t.chain,
+                t.chain_id,
+                CAST(t.contract_address AS String) AS contract_address,
+                t.num_transfers,
+                quantile({PERCENTILE})(t.num_transfers) OVER (PARTITION BY t.chain, t.chain_id) AS percentile_num_transfers
```
MSilb7, line 39: if this ever hits performance issues, I'm sure we can do something like "where transfers > 10" as well. But not necessary if it works fine.

src/op_analytics/datasources/chainsrpc/erc20metadata.py
```diff
@@ -0,0 +1,188 @@
...
+        function=lambda x: get_token_metadata(x),
+        targets=targets[:10],
+        max_workers=4,
+    )
+
+    return results
+
+
+def get_rpc_for_chain(chain_id: int) -> str:
+    df = load_chain_metadata()
```
MSilb7, line 70: If the public RPCs are ever not sufficient, we can store private RPCs

MSilb7 (approved):

src/op_analytics/datasources/chainsrpc/ddl/onchain.dim_erc20_token_metadata_v1.sql
lithium323: Having that `chain` string is super useful when debugging a table, so that's why I like to keep it around always.  

The `chain` string that we are using here is coming from the name of the table that we use in the Goldsky Clickhouse database, `oplabs_db_schema` in he [spreadsheet](https://docs.google.com/spreadsheets/d/1W_cp8po9bAU0VcnOmYnfQJ-D02hVXj2IS5ZpzMMKnmc/edit?gid=1407118431#gid=1407118431). 

It is a value that once picked we would never change. In other words it is not considered a display name, but rather a hard-coded value that just can't change (or would be a huge pain to change) after we start ingesting.  

If a chain's `chain_name` or `display_name` changes in our config we would still keep the original `oplabs_db_schema` value that we assigned to it.  Something that is confusing is we use that value as `chain` everywhere after ingestion.  Sometimes I wish I had used something other than `chain` post-ingestion.  To make it clear that this is how we internally refer to the chain in our pipelines and it is not likely to change.

lithium323:
* Added handling for some edge cases
* Added delays to avoid hitting the rate limits. The pull is slow but that's is ok because we are running it only once a day. 
* Since the pull is slow we are using clickhouse to store state, that way we can retry it safely.

Will go ahead and merge and open up a new pr for the scheduling changes.

PR #1375, opened 2025-02-19
net flow tested

chucktraverse:
Draft PR for net flows. Calculates net flows at the chain-protocol_slug-token level for the past 1, 7, 14, 28, 60, 90, and 365 days. 

Still left to do:
- test backfill and make temp table for team testing
- add in anomaly detection logic. this logic may be worth implementing earlier in the data pipelines. 
- anomaly detection looks for data anomalies that are x times greater than the previous and subsequent entires relative to a particular date. This helps reduce issues where a single bad value for a day can cause us to draw incorrect conclusions

### DIFF
--- a/src/op_analytics/datasources/defillama/net_tvl_flows.py
+++ b/src/op_analytics/datasources/defillama/net_tvl_flows.py
@@ -1,0 +1,137 @@
+from dataclasses import dataclass
+from datetime import timedelta
+
+import polars as pl
+
+from op_analytics.coreutils.duckdb_inmem.client import init_client
+from op_analytics.coreutils.logger import structlog
+from op_analytics.coreutils.partitioned.dailydatautils import dt_summary
+from op_analytics.coreutils.time import date_fromstr, now_dt
+from op_analytics.datasources.defillama.dataaccess import DefiLlama
+
+log = structlog.get_logger()
+
+FLOW_DAYS = [1, 7, 14, 28, 60, 90, 365]
+FLOW_TABLE_LAST_N_DAYS = 90
+
+
+def execute_pull():
+    # Produce the result
+    result = DefiLlamaNetFlows.of_date()
+
+    # Write to storage
+    DefiLlama.PROTOCOL_TOKEN_NET_FLOWS.write(
+        dataframe=result.df_net_flows,
+        sort_by=["dt", "chain", "protocol_slug", "token"],
+    )
+
+    return {
+        "df_net_flows": dt_summary(result.df_net_flows),
+    }
+
+
+@dataclass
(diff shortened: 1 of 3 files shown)


chucktraverse:
@lithium323 FYI I took the conscious approach to join on dt with two dfs rather than doing a dt shift on the tvl values directly. This avoids edge cases where a missing row causes a shift into the wrong date. It's rare but I was encountering this issue with polars for some reason. Joining on dt directly resolved this issue

src/op_analytics/datasources/defillama/net_tvl_flows.py
```diff
@@ -0,0 +1,137 @@
...
+        df_flows = calculate_net_flows(df_tvl, FLOW_DAYS)
+
+        return cls(df_net_flows=df_flows)
+
+
+def calculate_net_flows(df: pl.DataFrame, flow_days: list) -> pl.DataFrame:
+    # Ensure dt is a date. Adjust the parsing if your format differs.
+    # df = df.with_columns([pl.col("dt").str.strptime(pl.Date, fmt="%Y-%m-%d").alias("dt")])
+
+    # Sort by grouping keys and dt so that the join behaves as expected.
```
lithium323, line 80: this comment made me think, is there something in polars that make joins dependent on order? Or maybe this is more on the side of the `fill_null(0).fill_nan(0)` ?

src/op_analytics/datasources/defillama/net_tvl_flows.py
```diff
@@ -0,0 +1,137 @@
...
+            """
+        ).pl()
+
+        # Calculate net flows for each flow day period
+        df_flows = calculate_net_flows(df_tvl, FLOW_DAYS)
+
+        return cls(df_net_flows=df_flows)
+
+
+def calculate_net_flows(df: pl.DataFrame, flow_days: list) -> pl.DataFrame:
```
lithium323, line 76: Since the function has a straightforward df in -> df out signature it might be easy to create a sample dataframe in a test to check on the windowing logic.  This is something that cursor should be able to generate easily if you prompt it from a `tests/op_analytics/datasources/defillama/test_net_tvl_flows.py` file

MSilb7 (approved):
lfg

src/op_analytics/datasources/defillama/net_tvl_flows.py
chucktraverse: Yeah this is left over code I need to delete. It's from before I was doing the join methodology. I was doing .shift() which is dependent on dt order

lithium323:
Will go ahead and merge and will use another PR for any changes done while testing and backfilling.

PR #1380, opened 2025-02-20
Migrate teleportr to Clickhouse transforms

chuxinh:
Another attempt to migrate the underlying data for [Teleportr dashboard](https://app.hex.tech/61bffa12-d60b-484c-80b9-14265e268538/app/d7211b8f-75cb-4ecb-bd53-6e8bf63b053f/latest). 

Also realised across changed their log topic emitted from Feb 7 onwards so that we have not received any data since then.


src/op_analytics/transforms/teleportr/update/01_fact_teleportr_events_v1.sql
```diff
@@ -0,0 +1,131 @@
+INSERT INTO _placeholder_
+
+WITH across_bridge_metadata AS (
+    SELECT *
+    FROM dailydata_gcs.read_default(
+      rootpath = 'chainsmeta/raw_accross_bridge_gsheet_v1'
+    )
+  SETTINGS use_hive_partitioning = 1
```
lithium323, line 8: making a not to self to always provide this SETTING when passing the query to the clickhoue client so we don't have to manually have it in the SQL here. 

Also general comment that `SETTINGS use_hive_partitioning = 1` can be added only once at the very end of the .sql file and it will also work.  In other words, no need to add it on each subquery.

src/op_analytics/transforms/teleportr/update/01_fact_teleportr_events_v1.sql
```diff
@@ -0,0 +1,131 @@
...
+    , dt = { dtparam: Date }
+  )
+  WHERE true
+  AND topic0 in ('0xa123dc29aebf7d0c3322c8eeb5b999e859f39937950ed31056532713d0de396f', '0x32ed1a409ef04c7b0227189c3a103dc5ac10e775a15b785dcc510201f7c25ad3')
+  AND data is not null
+  AND data != ''
+  AND length(indexed_args) >= 3
+)
+, raw_transactions AS (
+  SELECT *
```
lithium323, line 32: It might be good to project only the columns you need from raw transactions here. That could help the optimizer to avoid reading everything from the parquet file.  And also makes it more clear what is needed after.

src/op_analytics/transforms/teleportr/update/01_fact_teleportr_events_v1.sql
```diff
@@ -0,0 +1,131 @@
...
+  ON cast(x.dst_chain_id as String) = cast(c.mainnet_chain_id as String)
+LEFT JOIN chainsmeta.dim_erc20_token_metadata_v1 AS mi
+  ON
+    x.input_token_address = mi.contract_address
+    AND cast(x.src_chain_id as String) = cast(mi.chain_id as String)
+LEFT JOIN chainsmeta.dim_erc20_token_metadata_v1 AS mo
+  ON
+    x.output_token_address = mo.contract_address
+    AND cast(x.dst_chain_id as String) = cast(mo.chain_id as String)
+WHERE integrator is not null
```
lithium323, line 131: should we push the `integrator is not null` filter up to the raw logs query?

lithium323: sorry I meant the raw transactions query

lithium323 (approved):

PR #1393, opened 2025-02-21
Implement sugar-sdk pool snapshot data pull

dioptx:
Implements the Sugar SDK puller that currently pulls snapshot data for the Base and OP chains.

### DIFF
--- a/src/op_analytics/datasources/sugar/chain_list.py
+++ b/src/op_analytics/datasources/sugar/chain_list.py
@@ -1,0 +1,4 @@
+from sugar.chains import BaseChain, OPChain
+
+# So far, sugar-sdk only supports BaseChain and OPChain
+chain_list = [BaseChain, OPChain]
--- a/src/op_analytics/datasources/sugar/chains/chains.py
+++ b/src/op_analytics/datasources/sugar/chains/chains.py
@@ -1,0 +1,60 @@
+from typing import Any, Dict, List, Tuple
+
+from op_analytics.coreutils.logger import structlog
+from op_analytics.datasources.sugar.prices.dynamic_prices import fetch_prices_with_retry
+
+log = structlog.get_logger()
+
+
+async def fetch_chain_data(
+    chain_cls: type,
+) -> Tuple[List[Dict[str, Any]], List[Dict[str, Any]], List[Dict[str, Any]]]:
+    """
+    Fetch chain data (tokens, pools, prices) for a given chain class.
+
+    Args:
+        chain_cls: A Sugar chain class (e.g. OPChain or BaseChain).
+
+    Returns:
+        A tuple with three lists:
+          - tokens_data: List of dictionaries for tokens.
+          - pools_data: List of dictionaries for liquidity pools.
+          - prices_data: List of dictionaries for token prices.
+    """
+    # Initialize chain instance (assumed async context manager)
+    async with chain_cls() as chain:
+        tokens = await chain.get_all_tokens(listed_only=True)
+        log.info(f"{chain_cls.__name__}: Fetched {len(tokens)} tokens.")
+
+        # Build token mapping if needed
+        tokens_data = [
+            {
+                "token_address": t.token_address,
+                "symbol": t.symbol,
+                "decimals": t.decimals,
+                "listed": t.listed,
+            }
+            for t in tokens
+        ]
+
+        pools = await chain.get_pools()
+        log.info(f"{chain_cls.__name__}: Fetched {len(pools)} liquidity pools.")
+        pools_data = [
+            {
+                "lp": p.lp,
+                "factory": p.factory,
+                "symbol": p.symbol,
+                "is_stable": p.is_stable,
+                "total_supply": p.total_supply,
+                "decimals": p.decimals,
+                "token0": p.token0.symbol if p.token0 else None,
+                "token1": p.token1.symbol if p.token1 else None,
+                "pool_fee": p.pool_fee,
+            }
+            for p in pools
+        ]
+
+        prices_data = await fetch_prices_with_retry(chain, tokens, initial_batch_size=40)
+        log.info(f"{chain_cls.__name__}: Fetched prices for {len(prices_data)} tokens.")
+
+    return tokens_data, pools_data, prices_data
--- a/src/op_analytics/datasources/sugar/execute.py
+++ b/src/op_analytics/datasources/sugar/execute.py
@@ -1,0 +1,67 @@
+"""
+Sugar data ingestion pipeline.
+
+This module pulls Sugar protocol data from all chains and writes the tokens, pools,
+and prices to partitioned datasets in ClickHouse or GCS (depending on the configuration).
+"""
+
+import polars as pl
+from typing import Dict, Any, List
+
+from op_analytics.coreutils.logger import structlog
+from op_analytics.coreutils.partitioned.dailydatautils import dt_summary
+from op_analytics.datasources.sugar.dataaccess import SugarDataAccess
+from op_analytics.datasources.sugar.chain_list import chain_list
+from op_analytics.datasources.sugar.chains.chains import fetch_chain_data
+
+log = structlog.get_logger()
+
+
+async def _collect_data() -> Dict[str, List[Dict[str, Any]]]:
+    """
+    Collects tokens, pools, and prices data from each configured chain (OPChain, BaseChain).
+    Returns:
+        Dictionary containing three lists, keyed by "tokens", "pools", and "prices".
+    """
+    all_data = {"tokens": [], "pools": [], "prices": []}
+
+    for chain_cls in chain_list:
+        tokens, pools, prices = await fetch_chain_data(chain_cls)
+        all_data["tokens"].extend(tokens)
+        all_data["pools"].extend(pools)
+        all_data["prices"].extend(prices)
+
+    return all_data
+
+
+def _write_data(
+    data: List[Dict[str, Any]],
+    dataset: SugarDataAccess,
+    data_type: str,
+) -> Dict[str, Any]:
+    """
+    Writes data to the dataset and returns a summary for logging.
+    """
+    df = pl.DataFrame(data)
+    dataset.write(df)
+
+    summary = {f"{data_type}_df": dt_summary(df)}
+    log.info(f"Sugar {data_type} ingestion completed", summary=summary)
+    return summary
+
+
+async def execute_pull() -> Dict[str, Any]:
+    """
+    Main Sugar ingestion entrypoint.
+    Fetches the data from all chains, writes to configured datasets,
+    and returns a summary dictionary.
+    """
+    all_data = await _collect_data()
+
+    summary: Dict[str, Any] = {}
+    summary.update(_write_data(all_data["tokens"], SugarDataAccess.TOKENS, "tokens"))
+    summary.update(_write_data(all_data["pools"], SugarDataAccess.POOLS, "pools"))
+    summary.update(_write_data(all_data["prices"], SugarDataAccess.PRICES, "prices"))
+
+    log.info("Sugar ingestion completed", summary=summary)
+    return summary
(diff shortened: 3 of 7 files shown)


src/op_analytics/datasources/sugar/chain_list.py
```diff
@@ -0,0 +1,4 @@
+from sugar.chains import BaseChain, OPChain
+
+# So far, sugar-sdk only supports BaseChain and OPChain
+chain_list = [BaseChain, OPChain]
```
lithium323, line 4: since this module will only ever have the chain list maybe we can declare it as a CHAIN_LIST constant in execute.py

src/op_analytics/datasources/sugar/chains/chains.py
```diff
@@ -0,0 +1,60 @@
...
+    Fetch chain data (tokens, pools, prices) for a given chain class.
+
+    Args:
+        chain_cls: A Sugar chain class (e.g. OPChain or BaseChain).
+
+    Returns:
+        A tuple with three lists:
+          - tokens_data: List of dictionaries for tokens.
+          - pools_data: List of dictionaries for liquidity pools.
+          - prices_data: List of dictionaries for token prices.
```
lithium323, line 22: Let's move the logic for each of these to the `tokens.py`, `pools.py`, and `prices.py`  modules. That way we can have all the fetching logic for each table more clearly isolated.

src/op_analytics/datasources/sugar/chains/chains.py
```diff
@@ -0,0 +1,60 @@
+from typing import Any, Dict, List, Tuple
+
+from op_analytics.coreutils.logger import structlog
+from op_analytics.datasources.sugar.prices.dynamic_prices import fetch_prices_with_retry
+
+log = structlog.get_logger()
+
+
+async def fetch_chain_data(
```
lithium323, line 9: we shouldn't use async functions in our part of the code. Since the sugar sdk is using async we can have a wrapper that lets use their coroutines as normal functions.  Async is difficult to reason about and we aren't really leveraging event loop concurrency here, so it's better to stay away from it. 

I'll open a PR to add a wrapper for the coroutines and will share it with you.

src/op_analytics/datasources/sugar/chains/chains.py
```diff
@@ -0,0 +1,60 @@
...
+                "token1": p.token1.symbol if p.token1 else None,
+                "pool_fee": p.pool_fee,
+            }
+            for p in pools
+        ]
+
+        prices_data = await fetch_prices_with_retry(chain, tokens, initial_batch_size=40)
+        log.info(f"{chain_cls.__name__}: Fetched prices for {len(prices_data)} tokens.")
+
+    return tokens_data, pools_data, prices_data
```
lithium323, line 60: general advice. If a function returns a tuple it usually means you need to break it apart into separate functions or if that's not possible then create a dataclass so you can return a single object with attribute names that callers can use to understand what is being returned.

src/op_analytics/datasources/sugar/execute.py
```diff
@@ -0,0 +1,67 @@
...
+async def execute_pull() -> Dict[str, Any]:
+    """
+    Main Sugar ingestion entrypoint.
+    Fetches the data from all chains, writes to configured datasets,
+    and returns a summary dictionary.
+    """
+    all_data = await _collect_data()
+
+    summary: Dict[str, Any] = {}
+    summary.update(_write_data(all_data["tokens"], SugarDataAccess.TOKENS, "tokens"))
```
lithium323, line 62: it might be simpler to update as:
```python
summary["tokens"] =  <... tokens write summary ...>
```

in other words you can nest the summaries, they don't need to be flat.  Summaries are only to get an quick idea of what happened from the dagster UI.

src/op_analytics/datasources/sugar/pools/pools.py
```diff
@@ -0,0 +1,57 @@
+from typing import List
```
lithium323, line 1: for the `pools/` and `prices/` modules maybe we don't need to make them directories, they can be `pools.py` and `prices.py`. It looks like there won't be a need to add more .py files in the directories later on.

src/op_analytics/datasources/sugar/pools/pools.py
```diff
@@ -0,0 +1,57 @@
...
+                "Fetched pool batch",
+                offset=offset,
+                batch_size=len(pools_batch),
+                total_pools=len(pools),
+            )
+            if len(pools_batch) < limit:
+                break
+            offset += limit
+
+        except Exception as exc:
```
lithium323, line 37: do you know what is the Exception class that gets returned on out of gas error? Catching that might be more robust than looking for the string in the exception message.

src/op_analytics/datasources/sugar/prices/dynamic_prices.py
```diff
@@ -0,0 +1,49 @@
...
+            log.info(
+                "Fetched token prices",
+                start=index,
+                end=index + current_batch_size,
+                total=len(tokens),
+            )
+            index += current_batch_size
+        except Exception as exc:
+            error_str = str(exc)
+            if "out of gas" in error_str or "0x3445e17c" in error_str:
```
lithium323, line 34: let's add more info about what the `0x3445e17c` error means

src/op_analytics/datasources/sugar/prices/dynamic_prices.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    Fetch prices for a list of tokens using dynamic batch sizing.
+
+    Retries with a reduced batch size if known errors occur.
+    """
+    prices = []
+    index = 0
+    current_batch_size = initial_batch_size
+
+    while index < len(tokens):
```
lithium323, line 20: here we could use `itertooles.batched()`

src/op_analytics/datasources/sugar/prices/dynamic_prices.py
```diff
@@ -0,0 +1,49 @@
+from typing import List, Any
+
+from op_analytics.coreutils.logger import structlog
+
+log = structlog.get_logger()
+
+
+async def fetch_prices_with_retry(
```
lithium323, line 8: the fault tolerance required to fetch tokens makes me think we can use the buffered approach we are using for defillama, so that we can recover from failures across separate runs on the same date.

lithium323 (changes requested):

MSilb7:
We can try more Chain IDs as well: https://github.com/velodrome-finance/sugar-sdk?tab=readme-ov-file#configuration

We store the OP Chain chain IDs in the chain metadata file (however is best to pull that). Not all will have Velodrome implementations. 

Also want to validate if we can pull things like votes, fees, etc before merging - since it's most valuable to have the full suite of metrics that you can pull from the UI (preferably at a token level for rewards and tvl).

callmephilip:
Howdy. We've just added some more [pool stats](https://github.com/velodrome-finance/sugar-sdk?tab=readme-ov-file#pools). This matches what velo/aerodrome display on liquidity screens. Do [let us know](https://github.com/velodrome-finance/sugar-sdk/issues) what other stats you might need. And thank you for your patience since all this is very much WIP at the moment

callmephilip:
`sugar-sdk 0.0.2` now supports both sync and async mode for existing readonly APIs. we are going to make sure that we support both modes moving forward. See docs: https://github.com/velodrome-finance/sugar-sdk?tab=readme-ov-file#pools

this will break your current setup since async chains must be prefixed with Async* now (e.g. OPChain became AsyncOpChain, its sync version is now called OpChain)

dioptx:
> `sugar-sdk 0.0.2` now supports both sync and async mode for existing readonly APIs. we are going to make sure that we support both modes moving forward. See docs: https://github.com/velodrome-finance/sugar-sdk?tab=readme-ov-file#pools
> 
> this will break your current setup since async chains must be prefixed with Async* now (e.g. OPChain became AsyncOpChain, its sync version is now called OpChain)

Thank you for flagging @callmephilip , we will review the new docs and reimplement.

src/op_analytics/datasources/velodrome/sugarwrapper.py
```diff
@@ -120,7 +113,7 @@ async def _sugar_pools(chain: str, sugar_chain: Chain) -> VelodromePools:
     lps = []
     missing = []
     for pool in pools:
-        lp = LiquidityPool.from_tuple(pool, tokens_index)
```
dioptx, line 123: @lithium323 

We have a bunch of missing Token information here still.

Sample data: 

```
Exception: possible error, token info not found for pools
[MissingTokenInfo(token0='0x4200000000000000000000000000000000000006', token1='0x557929D1d015F4B30E9C0E68cDF6e8a86B308E18', pool_type=-1),
 MissingTokenInfo(token0='0x2C00f7fF36c3591340b7DB47c22C4a413b925E5F', token1='0x4200000000000000000000000000000000000006', pool_type=-1),
 MissingTokenInfo(token0='0x4200000000000000000000000000000000000006', token1='0xF6A35217F35220efF1e3E1c489C30817245bc7cf', pool_type=-1),
 MissingTokenInfo(token0='0x4200000000000000000000000000000000000006', token1='0x92A062bB26A0Cbc704BF7dcD4C833d4e1beeb83d', pool_type=-1),
 MissingTokenInfo(token0='0x505269DA224Db7C3497A9799Fee06BbfbD0D40D2', token1='0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913', pool_type=-1),
 MissingTokenInfo(token0='0x06ef35365c3FA9125E94b4657E671FDaA0cB7B70', token1='0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913', pool_type=-1),
```

src/op_analytics/datasources/velodrome/sugarwrapper.py
```diff
@@ -120,7 +113,7 @@ async def _sugar_pools(chain: str, sugar_chain: Chain) -> VelodromePools:
...
+        lp = LiquidityPool.from_tuple(pool, tokens_index, prices)
```
dioptx, line 116: @lithium323 

We have a bunch of missing Token information for native tokens mostly. Not sure how to fix this yet.

Sample data: 

```
Exception: possible error, token info not found for pools
[MissingTokenInfo(token0='0x4200000000000000000000000000000000000006', token1='0x557929D1d015F4B30E9C0E68cDF6e8a86B308E18', pool_type=-1),
 MissingTokenInfo(token0='0x2C00f7fF36c3591340b7DB47c22C4a413b925E5F', token1='0x4200000000000000000000000000000000000006', pool_type=-1),
 MissingTokenInfo(token0='0x4200000000000000000000000000000000000006', token1='0xF6A35217F35220efF1e3E1c489C30817245bc7cf', pool_type=-1),
 MissingTokenInfo(token0='0x4200000000000000000000000000000000000006', token1='0x92A062bB26A0Cbc704BF7dcD4C833d4e1beeb83d', pool_type=-1),
 MissingTokenInfo(token0='0x505269DA224Db7C3497A9799Fee06BbfbD0D40D2', token1='0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913', pool_type=-1),
 MissingTokenInfo(token0='0x06ef35365c3FA9125E94b4657E671FDaA0cB7B70', token1='0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913', pool_type=-1),
```

lithium323: I brought this up with the velodrome team on telegram. At first setting listed_only=False made all pools with missing tokens go away.  We should check again with them if that isn't working anymore.

MSilb7:
Priorities to pull by pool:
- TVL
  - Breakdown by token if possible as well, but not ideal 
- Trade Volume
- veVELO votes
- Rewards (Fees + Incentives)
  - If this is by token, then this would be great to pull
  - Also, ideally we segment out fees from trade volume, versus fees from incentives/bribes.

If this is a super light pull, we can try to run ~hourly.

Also, if this shows data for each weekly epoch, then we could build a view on top that gets the amount that happens in each hour (i.e. if there's $100 volume now, $90 last hour, then $10 happened in the period elapsed).

PR #1424, opened 2025-02-28
Github metrics updates

lithium323:
1. The read_latest view was not robust with respect to pipeline delays. Fixed that by making it find the actual latest date. Can be better so we will look into improving that view.

2. Updating to polars 1.23 revealed we had some orphaned "_right" columns after joining that are not necessary so that gets fixed by using coalesce=True on the join.

**Description**

**Tests**

**Additional context**

**Metadata**


dioptx (approved):
LGTM

PR #1438, opened 2025-03-03
Defillama aggregated categories

chucktraverse:
uploads this sheet to BQ
https://docs.google.com/spreadsheets/d/1Iz-vHstZS7twRFSISQf9j5XMDyo_my27HY92p_C2998/edit?usp=sharing


lithium323 (approved):
looks good, my only question is if we can reuse the existing google sheet (use another tab) instead of creating a new google sheet. I think it would be a good idea to keep the metadata grouped in a single google sheet, but let me know if it's better to keep them separate.

lithium323 (approved):

PR #1453, opened 2025-03-05
Add daily dex trades pulls from dune

chuxinh:


src/op_analytics/datasources/dune/dataaccess.py
```diff
@@ -0,0 +1,14 @@
+from op_analytics.coreutils.logger import structlog
+from op_analytics.coreutils.partitioned.dailydata import DailyDataset
+
+log = structlog.get_logger()
+
+
+class Dune(DailyDataset):
```
lithium323, line 7: nice, this is a good one to have!

src/op_analytics/datasources/dune/dex_trades/daily_dex_trades_summary.py
```diff
@@ -0,0 +1,49 @@
+from dataclasses import dataclass
```
lithium323, line 1: I'm going to suggest we put this in `src/op_analytics/datasources/dune/dextrades.py`

lithium323 (approved):

PR #1457, opened 2025-03-06
Modify Flows Trailing Period

MSilb7:
Modifies the Flows logic to calculate ensuring we're inclusive of the current date (i.e. 7 day flow runs from today to today - 6 days, to get 7 total days)

### DIFF
--- a/src/op_analytics/datasources/defillama/protocolstvlenrich/read.py
+++ b/src/op_analytics/datasources/defillama/protocolstvlenrich/read.py
@@ -68,7 +68,8 @@ def read_lookback_data(compute_date: date, flow_days: list[int]):
 
     datevals = []
     for looback in flow_days:
-        datevals.append(compute_date - timedelta(days=looback))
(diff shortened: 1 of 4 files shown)


src/op_analytics/datasources/defillama/protocolstvlenrich/read.py
```diff
@@ -83,7 +84,7 @@ def read_lookback_data(compute_date: date, flow_days: list[int]):
         f"""
         SELECT
             dt,
-            DATE '{date_tostr(compute_date)}' - dt AS lookback,
+            DATE '{date_tostr(compute_date)}' - dt - 1 AS lookback,
```
lithium323, line 87: here you want to use +1 because the difference in the dates will be 6d but you want the number to be 7 since that `lookback` column value is used later on in determining column names

MSilb7:
Ignore, this issue was caused by the dashboard using an incomplete date to generate rolling prior periods.

PR #1459, opened 2025-03-07
Add governance tables using transforms approach

thbialek:
**Description**

**Tests**

**Additional context**

**Metadata**


src/op_analytics/transforms/governance/update/03_fact_votes_v1.sql
```diff
@@ -0,0 +1,31 @@
+/**
+
+Time series of votes on proposals.
+
+*/
+
+select
+    cast(t.block_timestamp as datetime) as dt
```
lithium323, line 8: I see this as a common pattern across all governance models. We should create a blocks fact table for the governance models that we can then reuse.  That way we can more easily add the timestamp for the blocks.

PR #1460, opened 2025-03-07
Create Transform for Dune Daily DEX Trades

chuxinh:
Create transform from GCS


src/op_analytics/transforms/dune/create/fact_daily_dex_trades_v1.sql
```diff
@@ -0,0 +1,35 @@
...
+    `project` String,
+    `version` String,
+    `token_pair_address_id` String,
+    `token_bought_address` String,
+    `token_sold_address` String,
+    `token_bought_symbol` String,
+    `token_sold_symbol` String,
+    `token_pair` String,
+    `sum_token_bought_amount` Float64,
+    `sum_token_bought_amount_raw` String,
```
lithium323, line 14: When writing it to Clickhouse you can convert these "raw" values to UInt256

PR #1471, opened 2025-03-12
Agged traces blockbatch model

lithium323:
**Description**

The tables produced here are necessary for daily trace aggregates.  At this level the aggregations are still pretty granular, so we run them in blockbatch mode for efficiency.  This is specially important since we have to join refined traces with refined transactions. 

**Tests**

**Additional context**

**Metadata**


src/op_analytics/datapipeline/models/templates/aggregated_traces/tr_from_tr_to_hash.sql.j2
```diff
@@ -0,0 +1,153 @@
...
+
+
+  , SUM(
+    (tx.l1_fee_native / tr.num_traces_in_txn)
+    + (tr.trace_gas_used - coalesce(tr.sum_subtraces_gas_used, 0)) * gwei_to_eth(tx.l2_gas_price_gwei)
+  ) AS sum_tx_fee_native_l1_amortized_l2_minus_subtraces
+
+
+  -- Count Call Types
+  , COUNT(*) FILTER (WHERE tr.trace_address = '' AND tr.call_type != '') AS count_top_level_calls
```
MSilb7, line 86: Maybe validate that NULL trace addresses are represented as empty string in our db?

PR #1604, opened 2025-05-09
Add daily transaction activity segments

dioptx:
Computes daily transaction activity segments in the blockbatchload daily pipeline


src/op_analytics/datapipeline/etl/blockbatchloaddaily/ddl/blockbatch_daily/wallet_segments/daily_tx_activity_segments_v1__INSERT.sql
```diff
@@ -0,0 +1,175 @@
...
+    -- Define segments based on daily transaction count
+    multiIf(
+      dwa.count_transactions >= 501, '>500_tx_per_day',    -- Tier5: Automated (High Freq)
+      dwa.count_transactions >= 51,  '51-500_tx_per_day',  -- Tier4: Automated (Low Freq)
+      dwa.count_transactions >= 16,  '16-50_tx_per_day',   -- Tier3: Power users
+      dwa.count_transactions >= 2,   '2-15_tx_per_day',    -- Tier2: Organic activity
+      dwa.count_transactions = 1,    '1_tx_per_day',       -- Tier1: Organic activity (Single)
+      'undefined' -- Fallback (shouldn't be hit due to WHERE clause above)
+    ) AS segment,
+    dwa.count_success_transactions / nullIf(dwa.count_transactions, 0) AS wallet_success_rate,
```
MSilb7, line 41: Is there a reason why we're adding these here? Should they be in the upstream model if we think it's important enough?

src/op_analytics/datapipeline/etl/blockbatchloaddaily/ddl/blockbatch_daily/wallet_segments/daily_tx_activity_segments_v1__INSERT.sql
```diff
@@ -0,0 +1,175 @@
...
+    AND count_transactions > 0
+),
+
+DailyWalletCohorts AS (
+  -- Assigns segments and calculates per-wallet rates/averages
+  SELECT
+    dwa.*, -- Select all columns from DailyWalletActivity
+    -- Define segments based on daily transaction count
+    multiIf(
+      dwa.count_transactions >= 501, '>500_tx_per_day',    -- Tier5: Automated (High Freq)
```
MSilb7, line 34: Maybe cut the tier definitions here, so that we're not locked in to them. 50+ doesn't necessarily mean automated, <15 doesn't necessarily mean organic

src/op_analytics/datapipeline/etl/blockbatchloaddaily/ddl/blockbatch_daily/wallet_segments/daily_tx_activity_segments_v1__INSERT.sql
```diff
@@ -0,0 +1,175 @@
...
+  avgWeighted(wallet_avg_l1_gas_per_tx, count_transactions) AS avg_wallet_avg_l1_gas_per_tx
+
+FROM DailyWalletCohorts
+-- Group by date, chain, network, and segment, AND also group by just date, chain, network (for baseline)
+GROUP BY GROUPING SETS (
+  (dt, chain, chain_id, network, segment),
+  (dt, chain, chain_id, network)
+)
+-- Order results for readability
+ORDER BY
```
MSilb7, line 162: Is this is unnecessary in production?

Also wonder if including the `multiIf` logic here as well will be hard to maintain vs 1,2,3,4

And should it be the `wallet_segment` field instead?

src/op_analytics/datapipeline/etl/blockbatchloaddaily/ddl/blockbatch_daily/wallet_segments/daily_tx_activity_segments_v1__INSERT.sql
```diff
@@ -0,0 +1,175 @@
...
+    avg_l1_blob_base_gas_price_gwei, sum_input_zero_bytes, sum_success_input_zero_bytes,
+    sum_input_bytes_length, sum_success_input_bytes_length, sum_input_nonzero_bytes,
+    sum_success_input_nonzero_bytes, sum_estimated_size, sum_success_estimated_size
+  FROM INPUT_CLICKHOUSE('blockbatch_daily/aggtxs/daily_address_summary_v1')
+  WHERE
+    1=1
+    AND count_transactions > 0
+),
+
+DailyWalletCohorts AS (
```
MSilb7, line 28: I wonder if `DailyWalletCohorts` should be a separate intermediate step?

Such that when we have many cohorts, we can throw them all at the same aggregations we have below?

i.e. DailyWaletCohorts UNION ALL DailySomethngCohorts -> Select from this in the last query?

MSilb7 (commented):
Added some ideas for if we wanted to set this up in a ~modular way to also handle for future cohorts.

But also open to making this one "just work" and we worry about making it modular when we do have another cohort to add.

dioptx:
Good point, I want to do something similar as a step two but trying to keep intermediate changes simple for now 🙏

PR #1612, opened 2025-05-13
Fix mypy

dioptx:


dioptx:
@MSilb7 Fixed a couple of mypy errors in some places, including unichain_lm

PR #1613, opened 2025-05-13
bump version

dioptx:


MSilb7 (changes requested):
Can we add the two new Dune pulls, then merge and update?

https://github.com/ethereum-optimism/op-analytics/blob/main/src/op_analytics/dagster/assets/dune.py
- eth_price_vol
- unichain_lm

PR #1621, opened 2025-05-19
Add multi chain segments

dioptx:


MSilb7 (approved):
Seems to make sense.

As a next PR, now that we have 2 segments in I wonder if we should try to "consolidate" some of the field definitions.

If a pattern becomes like:
- Fields to determine your segments
- [ All the same counting, summing, average, etc metrics ]

dioptx:
> Seems to make sense.
> 
> As a next PR, now that we have 2 segments in I wonder if we should try to "consolidate" some of the field definitions.
> 
> If a pattern becomes like:
> 
> * Fields to determine your segments
> * [ All the same counting, summing, average, etc metrics ]

Yes definitely, small steps but towards this direction 🚀

PR #1622, opened 2025-05-19
Add tx hash breakdown to trace-level agggregates

MSilb7:
Add tx hash breakdown to trace-level aggregates

Use case: When we want to pull metrics for transactions that involve calls to one of a set of trace to addresses (i.e. 100 oracle addresses), we need to be able to grab the unique transaction hashes before aggregating.


MSilb7:
Edit: May close this, for these use cases, we may be better doing it ad-hoc?

MSilb7:
> > Trying to backfill manually and some of the days are failing due to memory issues
> 
> Can you upload the backfill notebook as well? I can take a look.

It's the load_daily.ipynb, but not worth materializing. Leaving open until we talk through it though.

PR #1624, opened 2025-05-22
Fix L2beat pulls

dioptx:


chuxinh (approved):
LGTM. are you currently using these two utils anywhere? 

Thought we've been pulling l2beats data into pipeline as well

PR #1627, opened 2025-05-27
L2beat TVS fix

chuxinh:
Debug L2Beat tvs data pull


chuxinh:
More fix turns out to be needed as L2beat changed their project api structure 😮‍💨 
not pushing dagster image until we are good with the changes

PR #1629, opened 2025-06-02
ChainMetadataAggregator Integration

dioptx:
This is the first PR for the Total Crypto metadata integration. It provides the appropriate structure to base further additions and allows for smaller follow up PRs.

Additions:
- Created pipeline: build_all_chains_metadata() with 7 processing steps
- Dagster asset: all_chains_metadata_asset with config schema
- Config fields: output_bq_table, manual_mappings_filepath, bq_project_id, bq_dataset_id
- CLI command: build-metadata under opdata chains
- Registered Dagster asset module in defs.py

### DIFF
--- a/src/op_analytics/datapipeline/chains/aggregator.py
+++ b/src/op_analytics/datapipeline/chains/aggregator.py
@@ -1,0 +1,62 @@
+"""
+ChainMetadataAggregator module for op_analytics.datapipeline.chains.
+
+This module provides functionality to aggregate and process metadata from multiple chains,
+performing entity resolution, deduplication, and enrichment before outputting to BigQuery.
+"""
+
+from op_analytics.coreutils.logger import structlog
+
+log = structlog.get_logger()
+
+
+def build_all_chains_metadata(
+    output_bq_table: str,
+    manual_mappings_filepath: str,
+    bq_project_id: str,
+    bq_dataset_id: str,
+) -> None:
+    """
+    Build aggregated metadata for all chains with comprehensive processing pipeline.
+
+    This function orchestrates the complete chain metadata aggregation pipeline,
+    including data loading, preprocessing, entity resolution, deduplication,
+    enrichment, and output to BigQuery.
+
+    Args:
+        output_bq_table (str): Target BigQuery table name for aggregated metadata
+        manual_mappings_filepath (str): Path to manual mappings configuration file
+        bq_project_id (str): BigQuery project ID for data operations
+        bq_dataset_id (str): BigQuery dataset ID for table operations
+
+    Returns:
+        None
+    """
+    log.info("Pipeline execution started")
+
+    log.info(
+        "Pipeline configuration parameters",
+        output_bq_table=output_bq_table,
+        manual_mappings_filepath=manual_mappings_filepath,
+        bq_project_id=bq_project_id,
+        bq_dataset_id=bq_dataset_id,
+    )
+
+    log.info("Step 1: Load Data - Not yet implemented")
+    log.info("Step 2: Combine - Not yet implemented")
+    log.info("Step 3: Preprocess - Not yet implemented")
+    log.info("Step 4: Entity Resolution - Not yet implemented")
+    log.info("Step 5: Deduplication - Not yet implemented")
+    log.info("Step 6: Enrichment - Not yet implemented")
+    log.info("Step 7: Output - Not yet implemented")
+
+    log.info("Pipeline execution finished")
+
+
+if __name__ == "__main__":
+    build_all_chains_metadata(
+        output_bq_table="aggregated_chains_metadata",
+        manual_mappings_filepath="config/manual_mappings.json",
+        bq_project_id="op-analytics-dev",
+        bq_dataset_id="chains_metadata",
+    )
--- a/src/op_analytics/datapipeline/chains/aggregator.py
+++ b/src/op_analytics/datapipeline/chains/aggregator.py
@@ -20,8 +20,8 @@ def build_all_chains_metadata(
     Build aggregated metadata for all chains with comprehensive processing pipeline.
 
     This function orchestrates the complete chain metadata aggregation pipeline,
-    including data loading, preprocessing, entity resolution, deduplication,
-    enrichment, and output to BigQuery.
+    including data loading, preprocessing, combination, entity resolution, deduplication,
+    enrichment, validation, and output to BigQuery.
 
     Args:
         output_bq_table (str): Target BigQuery table name for aggregated metadata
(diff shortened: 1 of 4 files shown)


src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,62 @@
...
+        "Pipeline configuration parameters",
+        output_bq_table=output_bq_table,
+        manual_mappings_filepath=manual_mappings_filepath,
+        bq_project_id=bq_project_id,
+        bq_dataset_id=bq_dataset_id,
+    )
+
+    log.info("Step 1: Load Data - Not yet implemented")
+    log.info("Step 2: Combine - Not yet implemented")
+    log.info("Step 3: Preprocess - Not yet implemented")
```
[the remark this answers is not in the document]
MSilb7: I imagine like:
- We get data in from source A
- We modify the chain identifier in source A (but ideally this reads from some repeatable function, so if we change some logic it proliferates through to all sources)
- We can then use this modified version to combine

MSilb7 (commented):
Made one comment about maybe writing out what each step is supposed to do, so we can align on that before writing code.

otherwise, agree on a modular-type approach & the data sources go through a few steps before we create the massive joined dataset

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Step 1: Load Data
+    # Intent: Load chain metadata from multiple BigQuery data sources:
+    # - op_stack_chain_metadata from api_table_uploads (OP Labs source, source_rank=1)
+    # - Goldsky chain usage data from daily_aggegate_l2_chain_usage_goldsky (OP Labs source, source_rank=1)
+    # - L2Beat activity data from daily_l2beat_l2_activity (L2Beat source, source_rank=2)
+    # - L2Beat metadata extended from l2beat_metadata_extended (L2Beat source, source_rank=1.9)
+    # - GrowThePie activity data from daily_growthepie_l2_activity (GrowThePie source, source_rank=3)
+    # - Dune transaction data from dune_all_txs (Dune source, source_rank=4.5)
+    # - DefiLlama chain TVL data from daily_defillama_chain_tvl (DefiLlama source, source_rank=5)
```
MSilb7, line 55: I don't remember if I wound up using the ranks later on or not, but another note here is that we should be able to easily add any new datasource that comes in.

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+    # - Union all source_dfs into all_sources_df using pd.concat()
+    # - Ensure consistent schema across all sources before concatenation
+    # - Log total records combined from all sources
+    # - This creates the raw comprehensive dataset for further processing
+    log.info("Step 2: Combine - Not yet implemented")
+
+    # Step 3: Preprocess
+    # Intent: Clean and standardize the combined data with key transformations:
+    # - Generate chain_key column (hash of chain_name for grouping similar names)
+    # - Standardize chain_id column (ensure consistent format, handle special cases)
```
MSilb7, line 71: also just to note, we already know that chain_id collisions happen. So noting if this impacts the design at all.

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+    # - Apply data type conversions and handle missing values
+    # - Normalize chain names, symbols for consistency
```
MSilb7, line 73: this is where we can have some ~logic to pick the "best" Display Name (i.e. Arbitrum vs Arbitrum One, zkSync vs zkSync Era)

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+    # - This prepares data for entity resolution and grouping
+    log.info("Step 3: Preprocess - Not yet implemented")
+
+    # Step 4: Entity Resolution
+    # Intent: Generate unified_key and apply manual mappings before grouping:
+    # - Create unified_key column combining chain_id, chain_key, and display_name logic
```
MSilb7, line 79: this is the most difficult part.

So getting the earlier steps right helps.

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+    # - Within each group, sort by source_rank (lower rank = higher priority)
+    # - Select primary record from highest priority source
+    # - Aggregate fields: min_dt_day (min), max_dt_day (max), data_sources (concat), all_chain_keys (concat)
+    # - Apply secondary deduplication using is_dupe logic for remaining duplicates
+    # - Results in one canonical record per unique chain
+    log.info("Step 5: Deduplication - Not yet implemented")
+
+    # Step 6: Enrichment
+    # Intent: Enhance deduplicated data with op_stack_chain_metadata enrichment:
+    # - Load op_stack_chain_metadata specifically for enrichment (not as competing source)
```
MSilb7, line 98: Note for @chuxinh that whatever we generate out of registry ingestion + joining with our manual mappings will go here. Doesn't change anything, but just to show how it connects.

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+    # - Left join with deduplicated data on chain_id (handle Celo -l2 suffix matching)
```
MSilb7, line 99: The Celo handling was because the same chain switched over from an L1 to an L2. Ideally we handle this with some metadata file, rather than in the code.

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+    # - Coalesce/add fields: gas_token, da_layer, public_mainnet_launch_date, etc.
```
MSilb7, line 100: yeah these "added fields" could either be generated from the top, or we have some way to easily join the grouped chains back to their original datasets, such that we can pull the appropriate fields.

Like:
Chain A has aliases B for Dataset 1, C for Dataset 2, D for Dataset 3, etc and we left join back on that

src/op_analytics/datapipeline/chains/aggregator.py
```diff
@@ -0,0 +1,126 @@
...
+
+    # Step 7: Output
+    # Intent: Write the final enriched and deduplicated metadata to BigQuery:
+    # - Format final DataFrame according to target BigQuery schema requirements
+    # - Validate data quality and completeness before writing
+    # - Write to specified BigQuery table using write_full_df_to_bq with if_exists='replace'
+    # - Log success/failure status and record counts
+    # - Update processing metadata and create audit logs
+    log.info("Step 7: Output - Not yet implemented")
+
```
MSilb7, line 114: Maybe a step 8 (or 6.5?) can be some error checks. Like "Are there multiple rows with the exact same info, or like 80% of columns with the same info, etc.

PR #1645, opened 2025-06-10
Pipeline doc update

baokiddy:
**Description**

Updated docs adding more details on the pipeline deployment process

### DIFF
--- a/sphinx/devsetup/index.md
+++ b/sphinx/devsetup/index.md
@@ -48,6 +48,18 @@ The following directories/files are part of our legacy data platform and are sti
 - `value_locked_flows`
 - `Pipfile`
 
+For the deployment of new pipelines, the general flow of work should be as follows:
+- Proof of Concept notebook
+- An update secrets with latest credentials
+- Data backfill (as is needed/necessary)
+- Tests for final functional scripts
+- Update of the necssary Dagster and Kubernetes files
+- Run  `uv run mypy` to check the new python scripts
+- Run the script tests using `uv run pytest tests -vv`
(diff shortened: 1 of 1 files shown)


sphinx/devsetup/index.md
```diff
@@ -48,6 +48,18 @@ The following directories/files are part of our legacy data platform and are sti
 - `value_locked_flows`
 - `Pipfile`
 
+For the deployment of new pipelines, the general flow of work should be as follows:
+- Proof of Concept notebook
+- An update secrets with latest credentials
```
chuxinh, line 53: add in `.env` file locally?

dioptx: I would remove secret managment here as we document internally how it should be done and it's only done if needed, not on every pipeline development

sphinx/devsetup/index.md
```diff
@@ -130,19 +149,28 @@ version in two places in the codebase:
...
-You can search and replace for the version and replace it in both places at once. 
+You can search and replace for the version and replace it in both places at once.
 
 After updatatng the version then run the `make docker-dagster` command to build the image, tag it
-with the updated version and push it to the Github Container Registry. 
+with the updated version and push it to the Github Container Registry.
 
 Finally update the Dagster helm chart, telling it about the new image by running `make helm-dagster`.
 
 You can do both of the above at once with `make docker-dagster && make helm-dagster`.
```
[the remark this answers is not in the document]
dioptx: Yes, but it shouldn't be added here since the system prompts you to log in if you're not already logged in. It's a state, not a step per se.

PR #1648, opened 2025-06-17
Add Custom Gas Token - Daily Token Prices

MSilb7:
- Read custom gas token coingecko slugs (Eventually need to add a job that goes from CGT registry -> CG slug)
- Load daily prices in to our db

Still todo (need help):
- Update the dagster job

(Cursor did all of this, don't be impressed)


MSilb7:
merging so we can use the functions, will make a follow up PR for any edits/cleanup

PR #1682, opened 2025-07-11
Add mapping utils to support manual chain mappings

dioptx:
## Summary
Implements a configuration-driven manual mapping system to handle chain metadata transformations and data quality issues during the aggregation process.

## Changes

- Added: src/op_analytics/datapipeline/chains/mapping_utils.py - Core mapping system with validation and transformation logic
- Added: src/op_analytics/datapipeline/chains/resources/manual_chain_mappings.csv - Configuration file with mapping rules
- Added: tests/op_analytics/datapipeline/chains/test_mapping_utils.py - Comprehensive test suite
- Modified: src/op_analytics/datapipeline/chains/aggregator.py - Integration with mapping system

## Functionality

- Supports 6 mapping transformation types: chain_id_override, chain_id_suffix, display_name_preference, field_override, source_filter, conditional_transform
- CSV-based configuration with flexible rule definitions
- Source-specific filtering and conditional logic support
- Data validation and error handling
- Integration with existing polars and structlog patterns

## Testing

- Unit tests for all core functions
- Integration tests for complete rule application workflows
- Edge case handling for missing files and invalid rules

### DIFF
--- a/src/op_analytics/datapipeline/chains/mapping_utils.py
+++ b/src/op_analytics/datapipeline/chains/mapping_utils.py
@@ -1,0 +1,374 @@
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+import polars as pl
+
+from op_analytics.coreutils.logger import structlog
+
+log = structlog.get_logger()
+
+
+# Mapping rule types supported by the system
+SUPPORTED_MAPPING_TYPES = {
+    "chain_id_override",  # Override chain_id values (e.g., molten -> NULL)
+    "chain_id_suffix",  # Add suffix to chain_id (e.g., celo -> celo-l2)
+    "display_name_preference",  # Set preferred display names
+    "field_override",  # Override any field value
+    "source_filter",  # Apply rules only to specific sources
+    "conditional_transform",  # Apply rules based on conditions
+}
+
+# Valid identifier types for targeting records
+VALID_IDENTIFIER_TYPES = {
+    "chain_name",
+    "chain_id",
+    "display_name",
+    "source_name",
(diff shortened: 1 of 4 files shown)


src/op_analytics/datapipeline/chains/mapping_utils.py
```diff
@@ -0,0 +1,372 @@
+from pathlib import Path
```
MSilb7, line 1: is the right spot to handle for chain migration:
e.g., metadata only applies for a set date range.

Options:
- Start/end are null = don't care about dates
- Celo example: L1 ended on day X, L2 started on day Y
- Maybe we add options for block range and date range? (it depends how much granularity we have).

src/op_analytics/datapipeline/chains/mapping_utils.py
```diff
@@ -0,0 +1,372 @@
...
+
+import polars as pl
+
+from op_analytics.coreutils.logger import structlog
+
+log = structlog.get_logger()
+
+
+# Mapping rule types supported by the system
+SUPPORTED_MAPPING_TYPES = {
```
MSilb7, line 12: Tangible use case (if this handles for it):

- We have a chain id collision between shape and molten (360). Does this handle for fixing that (i.e. maybe we overwrite molten's 360 to 360-molten?)

src/op_analytics/datapipeline/chains/mapping_utils.py
MSilb7: could also be like "DA Layer migration" - I feel like migrations are the use case

PR #1728, opened 2025-08-26
Add jovian analysis

chuxinh:


notebooks/adhoc/jovian_analysis/jovian_src/core.py
```diff
@@ -0,0 +1,1121 @@
...
+        self.jovian_config = jovian_config
+        self.analysis_config = analysis_config
+        self.compressor = FastLZCompressor()
+
+    def calculate_size_estimate(self, calldata: bytes) -> float:
+        """Calculate Jovian size estimate for calldata."""
+        if not calldata:
+            return float(self.jovian_config.min_transaction_size)
+
+        fastlz_size = self.compressor.compress_len(calldata)
```
sebastianst, line 821: We forgot to add a constant `68` overhead to account for transaction metadata like signatures.
```suggestion
        fastlz_size = self.compressor.compress_len(calldata) + 68
```
E.g. see the [Fjord specs](https://specs.optimism.io/protocol/fjord/predeploys.html) where fastLZ was introduced, where we also account for this in the L1 gas estimation functions by adding `68` to the fastLZ length of calldata.

notebooks/adhoc/jovian_analysis/jovian_src/core.py
```diff
@@ -0,0 +1,1121 @@
...
+        """Analyze a single transaction."""
+        calldata_hex = tx_row.get('input', '0x')
+        calldata = parse_calldata(calldata_hex)
+
+        # OPTIMIZATION: Calculate FastLZ compression once and reuse for all calculations
+        # Original code had 3 redundant computations:
+        # 1. Direct compress_len() call
+        # 2. compress_len() inside calculate_size_estimate()
+        # 3. compress_len() inside calculate_footprint() -> calculate_size_estimate()
+        fastlz_size = self.compressor.compress_len(calldata)
```
sebastianst, line 843: same here
```suggestion
        fastlz_size = self.compressor.compress_len(calldata) + 68
```
wondering why the logic and calculation effort seem to be duplicated between `analyze_transaction` and `calculate_size_estimate`. The FastLZ compression is not super expensive but also not cheap to perform, so it would be better to deduplicate, also to make sure we have to define the size_estimate calculation only in one place.

notebooks/adhoc/jovian_analysis/jovian_src/core.py
```diff
@@ -0,0 +1,1121 @@
...
+
+class CalldataAnalyzer:
+    """Core analyzer for Jovian calldata footprints."""
+
+    def __init__(self, jovian_config: JovianConfig, analysis_config: AnalysisConfig):
+        self.jovian_config = jovian_config
+        self.analysis_config = analysis_config
+        self.compressor = FastLZCompressor()
+
+    def calculate_size_estimate(self, calldata: bytes) -> float:
```
sebastianst, line 816: May be a bit late in the game, but I realize that we took over the confusing term "size estimate" from Fjord. If everything `*size_estimate*` was called `*da_usage_estimate*` instead, it would better convey its meaning, since what we estimate is the usage of DA bytes of a transaction, or block. Size is ambiguous - later in the analysis we also now _estimate_ transaction and block _sizes_ by taking the calldata length and adding a constant 100.

Maybe an LLM can help us quickly do this rename. If not, no big deal.

PR #1799, opened 2026-07-08
chainsmeta: tolerate SystemConfig 4.0.0 removed getters

mds1:
SystemConfig 4.0.0 (op-contracts v8.0.0, https://github.com/ethereum-optimism/optimism/pull/21641) removes the `batchInbox()`, `startBlock()`, `BATCH_INBOX_SLOT()`, and `START_BLOCK_SLOT()` getters that the chainsmeta systemconfig ingest calls on every chain. Chains upgrade at different times, so the ingest must tolerate both versions indefinitely.

The pipeline already treats per-method reverts as None (that is how `operatorFeeScalar` and friends work on not-yet-upgraded chains), so reverts from removed getters were already handled. The actual gap: some nodes answer an `eth_call` to a removed method with an empty `0x` result instead of a revert, which previously produced garbage for address fields (`"0x0x"`) or raised for int fields. This PR treats an empty result the same as a revert, marks the two slot annotations `Optional`, and adds a test covering both failure shapes against a 4.0.0 SystemConfig.

No SQL changes — the ClickHouse columns are already `Nullable`. Historical rows keep their values; post-upgrade rows get NULL for the removed fields.

Tests: `uv run pytest tests/op_analytics/datasources/chainsmeta/ -v` → 19 passed; mypy and ruff clean.

Related PRs: https://github.com/ethereum-optimism/optimism/pull/21641 (contracts), https://github.com/ethereum-optimism/specs/pull/922 (spec), https://github.com/ethereum-optimism/superchain-ops/pull/1492 (upgrade tooling), the dev-docs edit rides along in the monorepo PR. Context: ethereum-optimism/optimism#21614.

### DIFF
--- a/src/op_analytics/datasources/chainsmeta/systemconfig/rpc.py
+++ b/src/op_analytics/datasources/chainsmeta/systemconfig/rpc.py
@@ -26,6 +26,19 @@ BLOCK_REQUEST = {
 
 
 # System Config Contract method IDs
+#
+# Not every method exists on every chain's SystemConfig, and chains upgrade at
+# different times, so we tolerate both old and new contract versions indefinitely:
+#
+# - Methods added in newer versions (e.g. operatorFeeScalar) revert on chains
+#   that have not upgraded yet.
+# - batchInbox, startBlock, BATCH_INBOX_SLOT, and START_BLOCK_SLOT were removed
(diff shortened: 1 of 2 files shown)


mds1:
Closing, this repo doesn't seem too maintained so I don't think we need this. However, if we actually still use this and you need help updating the contract ABIs let me know

